PR #4, opened 2012-12-18
PG support, plus fixes for rake

sax:
Here is some additional error handling to make makara work with the postgresql adapter.

We also found that some activerecord rake tasks asplode when you use any database adapter other than mysql, postgresql, sqlite or something called firebird. These are matched by duplicated case statements scattered throughout the activerecord rake files, and which are impossible to extend without hard-coding new adapters in.

We added some rake tasks which rewrite the activerecord configurations when running these AR tasks, to fake as if it is not there. This works for our deployment, but there may be other stray tasks which we missed.

### DIFF
--- a/lib/active_record/connection_adapters/makara_adapter.rb
+++ b/lib/active_record/connection_adapters/makara_adapter.rb
@@ -48,7 +48,7 @@ module ActiveRecord
       attr_reader :current_wrapper
 
       SQL_SLAVE_KEYWORDS      = ['select', 'show tables', 'show fields', 'describe', 'show database', 'show schema', 'show view', 'show index']
-      SQL_SLAVE_MATCHER       = /^(#{SQL_SLAVE_KEYWORDS.join('|')})/
+      SQL_SLAVE_MATCHER       = /^\s*(#{SQL_SLAVE_KEYWORDS.join('|')})/
 
       MASS_DELEGATION_METHODS = %w(reconnect! disconnect! reset!)
       MASS_ANY_DELEGATION_METHODS = %w(active?)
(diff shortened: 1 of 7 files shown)


mnelson:
Hey Eric / Paul, can you link to the places in AR where those case statements exist?

mnelson:
Nevermind, found some. What do you think about allowing the `config[adapter]` to be something like `"makara-mysql2"` which would load the makara adapter with the appropriate underlying adapter(s). This would also solve the rake task issues since they loosely check the adapter expression (`when /postgresql/`)

mnelson:
@evantahler what do you think about a config change https://github.com/taskrabbit/makara/compare/adapter_naming which addresses the issues brought on my case statements like: https://github.com/rails/rails/blob/ccf256d8dcc68d975bbf00780c3b4096f47e29d3/activerecord/lib/active_record/railties/databases.rake#L140

sax:
I think that since Rails 4 brings a more extendable syntax, where you can write your own adapter for rake tasks and register it, that the original pattern is superior. Hijacking rake tasks is very janky, but I think your approach of passing everything through to the original adapter is elegant and desirable.

sax:
By the way, we are about to launch a service written in Sinatra using Makara, which is why we switched the dependency from 'rails' to 'activerecord'. Everything works well so far, including running the test suite locally.

mnelson:
Hey guys, I just merged #7; please rebase master, remove the rake task stuff, get a green travis, and reopen this PR. I'll then review and merge the pg changes. Cheers

sax:
Awesome. I was out of town all last week, but I'll try to look at this as soon as I can.

PR #18, opened 2013-09-26
Rewrite

mnelson:
Makara 0.1.x was a database adapter. This is a generic solution which handles the management of N connections, potentially split between a master and slave group. The README.md has the basics.

Included is the ActiveRecord adapter implementation as well as Rack utilities.


mnelson:
@sax @kigster @evantahler @bleonard @davidjairala would really appreciate a review on this.

mnelson:
Might be easier to browse at this point: https://github.com/taskrabbit/makara/tree/rewrite

PR #22, opened 2013-11-07
Fixes double quotes for no ruby eval & cleans as per ruby style guide

yatish27:
- Removes double quotes for no ruby eval
- Adds space after comma
- Replace trivial methods with attr reader


mnelson:
Thanks

PR #31, opened 2014-01-09
Don't send locking sql to slaves

### LINKED ISSUE #30: .lock! method performed on slaves instead of master
We're doing database locking in one of our API callbacks, and this has broken since moving to makara. PostgreSQL handles locking using a SELECT FOR UPDATE a la `SELECT "video_representations".* FROM "video_representations" WHERE "video_representations"."id" = $1 LIMIT 1 FOR UPDATE`

[MakaraAbstractAdapter](https://github.com/taskrabbit/makara/blob/master/lib/active_record/connection_adapters/makara_abstract_adapter.rb#L52)'s regex doesn't catch this. I'm working on a fixed up regex right now but I'm a bit rusty in that department. failing that, i can work around this by doing a VideoRepresentation.touch or so to promote the request to master myself, but this is still, imo, a bug in the `#needs_master` method. If a single regex cannot cover this, a SQL_FORCE_MASTER regex which can catch SELECT FOR UPDATE coudl be inserted in to `#needs_master`, either way, I'm willing to submit the pull request for that. :)

mnelson:
Statements like the following were incorrectly going to slaves:

``` sql
SELECT * FROM USERS FOR UPDATE
SELECT * FROM USERS LOCK IN SHARE MODE
```

Should resolve #30


rrix-storehouse:
Tossing this ref at my master/slave testing environment right now

rrix-storehouse:
This looks to be fixing our issue! :+1:

PR #54, opened 2014-07-24
Require makara in makara abstract adapter

sax:
When using rails 4.1.2 release candidates, servers are unable to start because of
an error,

```
uninitialized constant Makara
from /path/to/bundle/ruby/2.1.0/bundler/gems/makara-59b01c0ecfec/lib/active_record/connection_adapters/makara_abstract_adapter.rb:4:in `<module:ActiveRecord>'
from /path/to/bundle/ruby/2.1.0/bundler/gems/makara-59b01c0ecfec/lib/active_record/connection_adapters/makara_abstract_adapter.rb:3:in `<top (required)>'
from /path/to/bundle/ruby/2.1.0/bundler/gems/makara-59b01c0ecfec/lib/active_record/connection_adapters/makara_postgresql_adapter.rb:1:in `<top (required)>'
```

I think that ActiveRecord must be requiring files in some slightly different order
than expected.

### DIFF
--- a/lib/active_record/connection_adapters/makara_abstract_adapter.rb
+++ b/lib/active_record/connection_adapters/makara_abstract_adapter.rb
@@ -1,4 +1,5 @@
 require 'active_record'
+require 'makara'
 
 module ActiveRecord
   module ConnectionAdapters


sax:
Nope. This was something we were doing with bundler in the effected application. This shouldn't actually be necessary.

neonishu:
@sax , Can you shed some light on the `something we were doing with bundler`, I am facing the same issue. I have an application serving APIs with Grape. And, this is the exact same issue that I faced. ActiveRecord version is `activerecord (4.2.7.1)` Appreciate a response here.

sax:
I wish I could help. This is so long ago, and I no longer work at / have access to the relevant repositories to look at git history.

neonishu:
cc/ @bleonard 
I am using Grape with Activerecord v 4.x. I am facing same issue, can we merge this pr/change ?
I tried using Makara version 0.4.0/0.3.10/0.3.9 but faced same issue with all of them.

bleonard:
Seems fine to me.

PR #57, opened 2014-09-23
Add CHANGELOG

mcansky:
In order to track a bit more easily what's happening in the project here is a CHANGELOG proposal.

I have started the history from 0.2.2 with what I could gather from the commit logs but more could certainly be added.

External reference for changelogs : http://keepachangelog.com/


mnelson:
Thanks, i'll add to it now.

mcansky:
@mnelson thanks !

PR #58, opened 2014-09-25
Fix for adapter name on readme

ravensnowbird:
The format is "makara_#{db_type}" although its a bit different for jdbcmysql_makara_adapter.rb and jdbcpostgresql_makara_adapter.rb

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,7 +144,7 @@ production:
         host: slave2.sql.host
 ```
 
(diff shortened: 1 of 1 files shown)


mnelson:
0.3.0.rc2 introduced the *_makara format. The makara_* is a remnant of the original implementation which I'll eventually remove. So there is an error in the readme, but it should just change mysql to makara.

PR #72, opened 2015-02-06
Handle leading spaces in the SQL

kdonovan:
In our project (postgres) the `needs_master?` method and friends were being passing SQL with leading spaces, which didn't match the regex and so _all_ queries ended up being routed to master.

### DIFF
--- a/lib/active_record/connection_adapters/makara_abstract_adapter.rb
+++ b/lib/active_record/connection_adapters/makara_abstract_adapter.rb
@@ -103,7 +103,7 @@ module ActiveRecord
       SQL_MASTER_MATCHERS           = [/^\s*select.+for update$/i, /select.+lock in share mode$/i].map(&:freeze).freeze
       SQL_SLAVE_MATCHERS            = [/^\s*select\s/i].map(&:freeze).freeze
       SQL_ALL_MATCHERS              = [/^\s*set\s/i].map(&:freeze).freeze
-      SQL_SKIP_STICKINESS_MATCHERS  = [/^\s*show\s([\w]+\s)?(field|table|database|schema|view|index)(es|s)?/i, /^(set|describe|explain|pragma)\s/i].map(&:freeze).freeze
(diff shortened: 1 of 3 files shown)


lib/active_record/connection_adapters/makara_abstract_adapter.rb
```diff
@@ -100,10 +100,10 @@ def custom_error_message?(connection, message)
       hijack_method :execute, :select_rows, :exec_query
       send_to_all :connect, :disconnect!, :reconnect!, :verify!, :clear_cache!, :reset!
 
-      SQL_MASTER_MATCHERS           = [/^select.+for update$/i, /select.+lock in share mode$/i].map(&:freeze).freeze
-      SQL_SLAVE_MATCHERS            = [/^select\s/i].map(&:freeze).freeze
-      SQL_ALL_MATCHERS              = [/^set\s/i].map(&:freeze).freeze
-      SQL_SKIP_STICKINESS_MATCHERS  = [/^show\s([\w]+\s)?(field|table|database|schema|view|index)(es|s)?/i, /^(set|describe|explain|pragma)\s/i].map(&:freeze).freeze
+      SQL_MASTER_MATCHERS           = [/^\s*select.+for update$/i, /select.+lock in share mode$/i].map(&:freeze).freeze
```
mnelson, line 103: There's a second regex here which should have the same matcher.

kdonovan, line 103: Good catch! One second...

mnelson:
I'm pretty sure the failure on travis is due to a new rack version if you want to update the httponly bit as well. Looks good after that.

kdonovan:
I tweaked the spec file to expect the HttpOnly, but now Travis is giving errors "Gem::InstallError: pg requires Ruby version >= 1.9.3.".  I didn't tweak the gemfile at all, so I'm not sure what's going on with that... both that and the spec issue seem outside the scope of the changes I made. :/

mnelson:
I'll take a look tomorrow.

PR #75, opened 2015-03-03
Work towards a fix for the Makara <> Newrelic issues.

mnelson:
Makara needs to be able to delagate to private methods but simpledelegate doesn't allow that. This branch is simply the source from simpledelegate with minor changes to allow private method forwarding.


joelr:
is this to address the performance issues?

mnelson:
yes

PR #76, opened 2015-03-31
Use same logging rules as ActiveRecord

ankane:
Hey, thanks for this awesome gem :+1:

This commit reduces logging noise by using [the same rules as ActiveRecord uses](https://github.com/rails/rails/blob/b06f64c3480cd389d14618540d62da4978918af0/activerecord/lib/active_record/log_subscriber.rb#L33).

### DIFF
--- a/lib/makara/logging/subscriber.rb
+++ b/lib/makara/logging/subscriber.rb
@@ -2,12 +2,17 @@ module Makara
   module Logging
 
     module Subscriber
+      IGNORE_PAYLOAD_NAMES = ["SCHEMA", "EXPLAIN"]
 
       def sql(event)
         name = event.payload[:name]
-        name = [current_wrapper_name(event), name].compact.join(' ')
-        event.payload[:name] = name
-        super(event)
+        if IGNORE_PAYLOAD_NAMES.include?(name)
+          self.class.runtime += event.duration
+        else
+          name = [current_wrapper_name(event), name].compact.join(' ')
+          event.payload[:name] = name
+          super(event)
+        end
       end
 
       protected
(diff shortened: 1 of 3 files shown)


lib/makara/logging/subscriber.rb
```diff
@@ -2,12 +2,17 @@ module Makara
...
     module Subscriber
+      IGNORE_PAYLOAD_NAMES = ["SCHEMA", "EXPLAIN"]
 
       def sql(event)
         name = event.payload[:name]
-        name = [current_wrapper_name(event), name].compact.join(' ')
-        event.payload[:name] = name
-        super(event)
+        if IGNORE_PAYLOAD_NAMES.include?(name)
+          self.class.runtime += event.duration
```
mnelson: Wouldn't you just want to call super here?

lib/makara/logging/subscriber.rb
```diff
@@ -2,12 +2,17 @@ module Makara
...
 
     module Subscriber
+      IGNORE_PAYLOAD_NAMES = ["SCHEMA", "EXPLAIN"]
 
       def sql(event)
         name = event.payload[:name]
-        name = [current_wrapper_name(event), name].compact.join(' ')
-        event.payload[:name] = name
-        super(event)
+        if IGNORE_PAYLOAD_NAMES.include?(name)
```
mnelson: This should just use `ActiveRecord::LogSubscriber::IGNORE_PAYLOAD_NAMES` directly if possible. Are there AR version semantics?

ankane: That's what I originally did, but `IGNORE_PAYLOAD_NAMES` isn't present before ActiveRecord 4 and could disappear at any time, so it's a bit safer to define it ourselves.  I don't feel too strongly either way.

lib/makara/logging/subscriber.rb
ankane: Good call.

PR #80, opened 2015-04-26
Improve set matcher

### LINKED ISSUE #70: Dealing with SET Edge cases
As README states: `The only edge case is SET operations which are sent to all connections.` Devise controller is using SET and thus leading to an error which one would like to avoid:
`PG::ReadOnlySqlTransaction: ERROR:  cannot execute UPDATE in a read-only transaction
: UPDATE "users" SET "last_sign_in_at" = '2014-12-09 19:58:58.290982', [...]`
notwithstanding the creation of an sql_proxy.rb initializer that should only handle select on the slave, as follows:

```
class MakaraSqlProxy < ::Makara::Proxy
  def connection_for(config)
    ::Sql::Client.new(config)
  end

  hijack_method :select, :ping
  send_to_all :connect, :reconnect, :disconnect, :clear_cache

  def needs_master?(method_name, args)
    return false if args.empty?
    sql = args.first
    sql !~ /^select/i
  end

end
```

righi:
fixes #70


mnelson:
I'm working on the travis build. Just ruby version & loose gem requirement issues.

PR #82, opened 2015-05-15
Makara::Proxy.new calls super to use SimpleDelegate

sax:
When ::ActiveRecord::Base.clear_active_connections! is executed in Rails 3,
it calls == on the connection. If initialize does not call super, the ==
method is not actually sent to the delegate, an ArgumentError: not delegated
error is raised.

Note that this does not happen in all instances, but is easily reproducible
when running Makara in the context of Sidekiq.

### DIFF
--- a/lib/makara/proxy.rb
+++ b/lib/makara/proxy.rb
@@ -49,6 +49,7 @@ module Makara
       @error_handler  ||= ::Makara::ErrorHandler.new
       @skip_sticking  = false
       instantiate_connections
+      super(config)
     end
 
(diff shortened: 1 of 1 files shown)


sax:
Rails 4 looks like it removed any == calls on connections when closing them. We're in progress on updating a Rails 3 app to Rails 4, starting by updating all of our gem dependencies.

mnelson:
Thanks!

mnelson:
v0.3.2 is up

PR #84, opened 2015-05-19
current and previous contexts are stored in a thread local variable.

mattcamuto:
context is now thread safe. can be used in a system like sidekiq

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,9 @@ ctx = Makara::Context.generate # or any unique sha
 Makara::Context.set_current ctx
 ```
 
+A context is local to the curent thread of execution. This will allow you to stick to master safely in a single thread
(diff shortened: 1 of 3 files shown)


mnelson:
Looks like we need to check for the existence of the Thread#thread_variable_set and backup to Thread.current#[] if it's not present.

mattcamuto:
@mnelson Just saw this. We will make these changes soon to be compatible pre 2.0

mnelson:
Great, thanks. Everything else looks good.

mattcamuto:
@mnelson I think this is good to go all is green. Please accept if all good.

mnelson:
Thanks so much. Merged and up at https://github.com/taskrabbit/makara/releases/tag/v0.3.3

Issue #91, opened 2015-08-25
DTrace probes?


sax:
I have some DTrace probes added to a branch I created to test out #87 in our infrastructure. Would you be interested in having a pull request for these?

I can understand your response either way... Feature utility vs. feature maintenance. Just let me know!

bleonard:
hey @sax - a few thoughts.
maybe we could generically hook those spots you found interesting in a callback or something. then people could do what they wanted. `analytics_handler` or something. for example if you just said `dtrace` maybe that ones built in but you could also give a proc or a class name.

I'm thinking there are lots of spots to allow named or generic callbacks like that. one case is from https://github.com/taskrabbit/makara/issues/85 where there is a way to have an override the pool selection.

@mnelson suggested 

```
master_node_selector: 'Makara::NodeSelectors::SingleWithFailover'
```

sax:
This sounds like a good solution. I'll work on a pull request now.

bleonard:
@sax @jacobbednarz any thoughts on these two PRs (https://github.com/taskrabbit/makara/pull/126 and https://github.com/taskrabbit/makara/pull/103)?

I'm just catching up and attempting to be a good citizen here. Are you using either one of them in prod?

sax:
I'm no longer at the company where we did this work. They are still using #103 in production, but I don't see any problems with going directly to ActiveSupport::Notifications (unless you want to eventually drop the hard dependency on ActiveSupport). It would be pretty easy to switch over the DTrace probes to hook into AS notifications rather than the custom notification module.

#126 doesn't provide any tests for the notifications, if you care about such things.

bleonard:
Thanks @sax

jacobbednarz:
The purpose behind my pull request was getting some sort of framework in
place so that we could monitor behaviour _before_ launching it into
production. Due to the scale this was going to be used on, I wasn't
comfortable using this without some instrumentation that we could
benchmark against our existing stack and confirm this didn't introduce
performance regressions. 

I will be revisiting the read/write split for our DB's however at this
stage, it won't be in the immediate future so you're free to do whatever
you'd like with the pull request. You're more than welcome to work on
that branch if you'd like to continue it.

metatron1973:
Instacart system is embedded with fraud inside the whole system how drivers are rated and tips being paid .....

PR #103, opened 2016-01-06
Instrument Cache, Context and Proxy with notification callbacks

sax:
Code fires notifications using a Makara::Notifications module. Users may register arbitrary callbacks for these instrumentation points.

### DIFF
--- a/lib/makara/cache.rb
+++ b/lib/makara/cache.rb
@@ -16,10 +16,13 @@ module Makara
       end
 
       def read(key)
-        store.try(:read, key)
+        store.try(:read, key).tap do |value|
+          Notifications.notify!('Cache:read', key, value)
+        end
       end
 
       def write(key, value, ttl)
+        Notifications.notify!('Cache:write', key, value, ttl)
         store.try(:write, key, value, :expires_in => ttl.to_i)
       end
 
--- a/lib/makara/notifications.rb
+++ b/lib/makara/notifications.rb
@@ -1,0 +1,40 @@
+module Makara
+  # A notification handler that allows users to register arbitrary instrumentation callbacks. For instance, this can be
+  # used to fire dtrace or strace probes, or to write log messages. Callbacks are in the form of blocks, which can take
+  # 0 to N arguments, where N is the number of arguments provided to the the :notify!: method.
+  #
+  # Example:
+  #
+  #   Makara::Cache instruments the :write: method as follows:
+  #
+  #     Notifications.notify!('Cache:write', key, value, ttl)
+  #
+  #   Callbacks can be registered as follows:
+  #
+  #     Notifications.register_callback('Cache:write') do |key|
+  #       puts "Makara::Cache.write: #{key}"
+  #     end
+  #
+  #     Notifications.register_callback('Cache:write') do |key, value|
+  #       puts "Makara::Cache.write: #{key}: #{value}"
+  #     end
+  #
+  module Notifications
+    class << self
+      def register_callback(name, &blk)
+        registered_callbacks[name] << blk
+      end
+
+      def notify!(name, *args)
+        return unless registered_callbacks[name]
+        registered_callbacks[name].each do |callback|
+          callback.call(*args)
+        end
+      end
+
+      def registered_callbacks
+        @registered_callbacks ||= Hash.new([])
+      end
+    end
+  end
+end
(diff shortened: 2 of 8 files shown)


sax:
This should solve #91.

lib/makara/cache.rb
```diff
@@ -16,10 +16,13 @@ def store=(store)
       end
 
       def read(key)
-        store.try(:read, key)
+        store.try(:read, key).tap do |value|
```
bleonard, line 19: why do some have the `tap` pattern and others do it above/below?

lib/makara/notifications.rb
```diff
@@ -0,0 +1,40 @@
+module Makara
+  # A notification handler that allows users to register arbitrary instrumentation callbacks. For instance, this can be
+  # used to fire dtrace or strace probes, or to write log messages. Callbacks are in the form of blocks, which can take
+  # 0 to N arguments, where N is the number of arguments provided to the the :notify!: method.
+  #
+  # Example:
+  #
+  #   Makara::Cache instruments the :write: method as follows:
+  #
+  #     Notifications.notify!('Cache:write', key, value, ttl)
```
bleonard, line 10: does your dtrace implementation have to know about all the keys? Think there is a use case for subscribing to `*`?

lib/makara/cache.rb
sax, line 19: I only use `tap` when there is data that I want to instrument that is not directly accessible as a one-liner. For instance, with `#write` all of the values are there from the method signature. With `#read`, the key is there from the method signature, but value is obtained from the cache store.

The alternative would be:

``` ruby
value = store.try(:read, key)
Notifications.notify!('Cache:read', key, value)
value
```

bleonard:
this might be another way of doing it.
more or less the same thing using the ActiveSupport pattern.
We already have that dependency bc of ActiveRecord. Might be worth using if it is stable API.

http://metaskills.net/2013/12/15/instrumenting-your-code-with-activesupport-notifications/
http://edgeguides.rubyonrails.org/active_support_instrumentation.html

lib/makara/cache.rb
bleonard, line 19: I see. cool.
Nice to understand the approach. I'm super old school, but over all the gem is all ruby-fancy so I can get behind it :-)

lib/makara/notifications.rb
sax, line 10: With dtrace, USDT probes need to be registered with specific argument types. For instance, this is how we have been doing things: https://github.com/wanelo/makara/blob/dtrace/lib/makara/dtrace_provider.rb. It calls through to C bindings, so if that is not done correctly the process segfaults. _After_ probes are registered, you can subscribe to them on the command line to make them actually emit information, so that's where I would listen to the `ruby*:makara::`

DTrace is fairly low-level, and you generally need to know how many arguments there are as well as the argument type. I'm not sure if other dynamic tracing libraries are friendlier on that front.

sax, line 10: Or rather, if you register bindings and then pass them incorrect data, your process can segfault.

sax:
Oh, yeah... I forgot about those. Do you always have ActiveSupport, or is that just when using Makara with ActiveRecord? I've never tried to use it outside of Rails.

bleonard:
currently activerecord is a hard dependency.
and just noticed that `proxy.rb` is doing this.

```
require 'active_support/core_ext/class/attribute'
require 'active_support/core_ext/hash/keys'
```

so it's not any worse, as they say. might be worth it for standardization and less code overall.

jacobbednarz:
I've just stumbled on this one and I'd definitely +1 a move to ActiveSupport - especially considering it's already a dependency for another class. Using DTrace probes is quite a specific way to handle this and might be outside the realm of most people touching this library. ActiveSupport is quite flexible and is able to be utilised across a range of tools (depending on how you choose to listen for it) so I think that will still enable you get the events you need from the base library and then in your application specific code, drop in DTrace/statsd/whatever calls for logging.

jacobbednarz:
The purpose behind my pull request was getting some sort of framework in
place so that we could monitor behaviour _before_ launching it into
production. Due to the scale this was going to be used on, I wasn't
comfortable using this without some instrumentation that we could
benchmark against our existing stack and confirm this didn't introduce
performance regressions. 

I will be revisiting the read/write split for our DB's however at this
stage, it won't be in the immediate future so you're free to do whatever
you'd like with the pull request. You're more than welcome to work on
that branch if you'd like to continue it.

sax:
Closing because I haven't been using Ruby in the last few years and am not sure if these changes are still relevant.

Issue #99, opened 2015-12-10
Replication lag


dreyks:
Everywhere I read about Makara I see a line regarding master stickiness countering the replication lag. But I can't quite get this thing to work. 

Judging from the code, Makara writes current context with needed ttl to cache in case this context stuck to master. That's ok. But then to know what the previous context was, a cookie is used, and this cookie lives only for 5 seconds, and this time is not affected by `master_ttl` config option. What if my replication lag is bigger? Say I put `master_ttl: 20` for a 20 seconds master stickiness, but the cookie still expires in 5 seconds, so Makara generates a new "previous" context, that doesn't match the context stored in cache.

Am I doing something wrong?

evantahler:
The way it is now, `master_ttl` is a setting which can be thought of per thread, and is invoked in the `Proxy`.  It's used in the `stick_to_master!` method indicating that a certain request/worker/etc should remain on master for the duration of the request.  We actually use `Thread.current.object_id` under the hood here.  This is used, for example, in a Rails request you create a new object and you need to read it back.  This is one blocker for "internal" replication delays.

Separate to this, the makara cookie is used to tell the client to inform the server that in a subsequent request that the web server should stick to master for duration of the request. This can inform the `Proxy` to `stick_to_master`.  As you point out, this cookie TTL is currently hard-coded to be [5s](https://github.com/taskrabbit/makara/blob/master/lib/makara/middleware.rb#L103).  

I agree that this should be configurable.

dreyks:
I see. But the question still stands: How do I counter replication lag of
20 seconds?

evantahler:
A possible solution is being discussed here: https://github.com/taskrabbit/makara/pull/100

bleonard:
Before, makara would only use it for the next request. https://github.com/taskrabbit/makara/pull/194 should fix this so it can be 20 seconds or more.

PR #108, opened 2016-01-29
Add specs

anthonyalberto:
Make existing ones pass

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,4 +15,3 @@ gem 'pg', :platform => :ruby
 
 gem 'activerecord-jdbcmysql-adapter', :platform => :jruby
 gem 'activerecord-jdbcpostgresql-adapter', :platform => :jruby
-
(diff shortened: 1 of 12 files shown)


anthonyalberto:
Whoops, disregard, was intended to be a PR on our fork :)

bleonard:
anything that should be contributed back?

anthonyalberto:
It's probably a bit rough given that some stuff has been tailored to our needs but I'll list what was done there : 
- Let the same user be able to do more than one http request that sticks to master (until the end of ttl). We had problems with ajax queries that would just make the next query stop sticking to master.
- Send back a custom http header as well as the cookie for mobile app support
- Implement `force_slave` and `force_master` methods that allows at any point in time to bypass the `needs_master?` logic and just force the pool. Sometimes we just need to send specific queries to slave.
- Add a Sidekiq middleware that cleans up context between jobs
- Stop using a server-side data-store and just rely on cookie/header for the TTL. Value of cookie now therefore includes the TTL.

Anything in particular that would be of interest?

bleonard:
The cookie thing (`stick_to_master_until`) be relevant based on what I've heard. For example, this issue: https://github.com/taskrabbit/makara/issues/99

PR #116, opened 2016-05-20
Add postgis support

choubacha:
Wraps the postgis adapter in the Makara abstract class.

This is only tested against AR 4.2+ and only on the ruby platform.


bleonard:
included in release candidate. https://github.com/taskrabbit/makara/releases/tag/v0.3.7.rc

PR #96, opened 2015-11-17
transaction support

bleonard:

PR #120, opened 2016-06-15
Choubacha/add tests for polymorphic associations

choubacha:
FYI: This should have failing tests for the postgresql adapter.

Added specs that reproduce an issue facing transactions. Once a transaction has been opened it would make sense to stick to master for the duration of the transaction. Otherwise selects will not be able to see any of the just updated or inserted or deleted records. It should be noted that tests surrounding the find or count method **do** pass.

If `sticky: false` is set in the database.yml then executes and polymorphic associations appear to hit the slave.

### DIFF
--- a/spec/support/schema.rb
+++ b/spec/support/schema.rb
@@ -1,9 +1,9 @@
-ActiveRecord::Schema.define(:version => 20130628161227) do
-  create_table "users", :force => true do |t|
-    t.string   "name"
+ActiveRecord::Schema.define(version: 20130628161227) do
+  create_table :users, force: true do |t|
+    t.string   :name
   end
 
-  create_table :pictures, :force => true do |t|
+  create_table :pictures, force: true do |t|
     t.string  :name
     t.integer :imageable_id
     t.string  :imageable_type
--- a/spec/support/schema.rb
+++ b/spec/support/schema.rb
@@ -1,7 +1,14 @@
 ActiveRecord::Schema.define(:version => 20130628161227) do
-
   create_table "users", :force => true do |t|
     t.string   "name"
   end
 
+  create_table :pictures, :force => true do |t|
+    t.string  :name
+    t.integer :imageable_id
+    t.string  :imageable_type
+    t.timestamps null: false
+  end
+
+  add_index :pictures, :imageable_id
 end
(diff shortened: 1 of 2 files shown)


spec/support/schema.rb
```diff
@@ -1,7 +1,14 @@
 ActiveRecord::Schema.define(:version => 20130628161227) do
-
   create_table "users", :force => true do |t|
     t.string   "name"
   end
 
+  create_table :pictures, :force => true do |t|
+    t.string  :name
```
jcraigk: Nitpick: 1.9 hash syntax

spec/support/schema.rb
```diff
@@ -1,7 +1,14 @@
 ActiveRecord::Schema.define(:version => 20130628161227) do
-
   create_table "users", :force => true do |t|
     t.string   "name"
   end
 
+  create_table :pictures, :force => true do |t|
```
jcraigk: Nitpick: 1.9 hash syntax

jcraigk:
lgtm_wonka

bleonard:
for better or worse, it looks to me like `sticky: false` explicitly says _not_ to stick to master (for reads) even after a write. 

Now, that might be the wrong thing. People might want that preference to only work before a write. Or maybe there should be two preferences. But eaither way, that's how it's coded up at the moment.

Any reason you can't switch that `sticky: true` ?

choubacha:
@bleonard 

Yea, Some of the code is a little confusing in there but I believe that there is _some_ built in stickiness for transactions. I know this because these two tests pass:

https://github.com/taskrabbit/makara/pull/120/files#diff-4804facb114de35b2faf60bce399d4acR56
https://github.com/taskrabbit/makara/pull/120/files#diff-4804facb114de35b2faf60bce399d4acR62

And we tested `sticky: false` before deployment and noticed that it was sticky within a transaction for general purpose queries. Perhaps this is something that rails does on it's own but without stickiness inside transaction there are severe issues (mainly that a write cannot be read back).

We tried to turn stickiness on, but it appears to hold over for up to 5 seconds and pegs all subsequent reads (even those in a read-only route). I think this has to do with the concept of contexts and how stickiness is used between contexts. Because of that we ended up seeing a CPU/IO profile that seemed to send nearly all reads/writes to master once it was turned on.

bleonard:
moving this here: https://github.com/taskrabbit/makara/pull/96

PR #121, opened 2016-07-09
pass class reference to middleware builder

dreyks:
Passing strings or symbols to the middleware builder is deprecated


bleonard:
probably replaced with this one: https://github.com/taskrabbit/makara/pull/122

PR #132, opened 2016-09-13
allow application configurable option to skip middleware cookie support

grossjo:
as well setup for future configurable options

### DIFF
--- a/lib/makara/railtie.rb
+++ b/lib/makara/railtie.rb
@@ -1,8 +1,17 @@
 module Makara
   class Railtie < ::Rails::Railtie
 
-    initializer "makara.configure_rails_initialization" do |app|
-      app.middleware.use Makara::Middleware
+    config.makara = ActiveSupport::OrderedOptions.new # enable namespaced configuration in Rails environments
+
+    initializer "makara.configure" do |app|
+      Makara.configure do |config|
+        config.skip_middleware = false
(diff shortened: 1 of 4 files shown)


grossjo:
Let me know if this interests you.
My application doesn't have support for cookies from clients, so the middleware is pointless for me.

It also, opens up the ability to pass in procs, in the config, as I saw you had a todo around that.

lib/makara/railtie.rb
```diff
@@ -1,8 +1,17 @@
...
+
+    initializer "makara.configure" do |app|
+      Makara.configure do |config|
+        config.skip_middleware = false
+        if app.config.makara[:skip_middleware]
+          config.skip_middleware = app.config.makara[:skip_middleware] 
+        end
+      end
+
+      app.config.middleware.insert_before "::Rails::Rack::Logger", "Makara::Middleware"
```
jeremy, line 14: Why install the middleware at all if we're going to skip it? Can we omit it entirely?

jeremy:
(Alternatively, you can `config.middleware.delete ::Makara::Middleware` in an initializer.)

bleonard:
I realize the response time is not so great, but let me know if this is needed (rather than just omitting it).

bleonard:
thanks for your interest @grossjo - let me know how it is working for you.

PR #135, opened 2016-09-27
Make makara error base class that inherits active record error class

bf4:
We noticed that Makara was raising exceptions that inherit from StandardError.  We thought it might be a better pattern for Makara to have its own error class that each error inherits from, and then for Makara's base error to inherit from an ActiveRecord error.  I chose the lowest level one assuming it is minimally acceptable.

Now code that was

``` ruby
rescue StandardError => e
  if e.class.name =~ /^Makara::/
```

can be

``` ruby
rescue Makara::Errors::MakaraError
```

### DIFF
--- a/lib/makara.rb
+++ b/lib/makara.rb
@@ -12,6 +12,7 @@ module Makara
   autoload :Proxy,              'makara/proxy'
 
   module Errors
+    autoload :MakaraError,                'makara/errors/makara_error'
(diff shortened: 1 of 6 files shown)


bf4:
@bleonard thoughts? anything I can help with?

bleonard:
lgtm! thanks 👍

Issue #138, opened 2016-10-14
API Rails App breaks Makara with assets.prefix no method error


will-in-wi:
When I create an API application which doesn't have Sprockets, I get the following error on page load:

```
#<NoMethodError: undefined method `assets' for #<Rails::Application::Configuration:0x007fb21491cf80>>
```

Adding the following to `config/application.rb` hacks around it for me:

```
config.assets = Struct.new('AssetsStub', :prefix).new('/assets')
```

FYI

bleonard:
https://github.com/taskrabbit/makara/pull/139

PR #139, opened 2016-10-15
check if using assets

bleonard:
https://github.com/taskrabbit/makara/issues/138

### DIFF
--- a/lib/makara/middleware.rb
+++ b/lib/makara/middleware.rb
@@ -37,7 +37,7 @@ module Makara
     # ignore asset paths
     # consider allowing a filter proc to be provided in an initializer
     def ignore_request?(env)
-      if defined?(Rails)
+      if defined?(Rails) && Rails.try(:application).try(:config).try(:assets).try(:prefix)
         if env['PATH_INFO'].to_s =~ /^#{Rails.application.config.assets.prefix}/
           return true
         end


lib/makara/middleware.rb
```diff
@@ -37,7 +37,7 @@ def call(env)
     # ignore asset paths
     # consider allowing a filter proc to be provided in an initializer
     def ignore_request?(env)
-      if defined?(Rails)
+      if defined?(Rails) && Rails.try(:application).try(:config).try(:assets).try(:prefix)
```
jeremy, line 40: Will `Rails.application` ever be missing? Think we can do `Rails.application.config.assets.try(:prefix)` since `config.assets` is the only possibly-nil bit.

bleonard: This gem works in Sinatra an other Rack apps but we added Rails checks for that common case.

jeremy:
Ah gotcha! Rails is defined in those contexts too?

bleonard:
No, it won't be defined and this method will return false in a Rack app and it will not ignore the request. We never had an Rack apps that had assets (or other requests to ignore) so I guess it never was an issue.

bleonard:
Thanks for reminding me this existed @jeremy - merged.

PR #141, opened 2016-10-19
Addressing the Rails 5 Question

that-jill:
On a fresh checkout of makara, bundler will attempt to install the latest version of ActiveRecord (currently > 5.0), but this will fail because the default ruby version (in `.ruby-version`) is too low.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ gemfile:
   - gemfiles/ar40.gemfile
   - gemfiles/ar41.gemfile
   - gemfiles/ar42.gemfile
+  - Gemfile
 
 sudo: false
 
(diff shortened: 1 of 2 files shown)


that-jill:
I understand that this isn't a long-term solution, but this demonstrates the issues for someone starting to work on makara. @bleonard do you have any thoughts on longer-term solutions for what AR support should be?

jeremy:
I'd suggest always having Makara master tracking Active Record master. Easy to slip into regressions and other unwelcome surprises, otherwise! So it'd make sense to bump .ruby_version instead and update the Travis matrix to be selective about which combinations of Ruby and AR to test.

bleonard:
I believe this is fixed now with the travis fixes in https://github.com/taskrabbit/makara/pull/160

bleonard:
@that-jill please let me know how things are going and sorry for the pitiful response time.

PR #143, opened 2016-12-07
Use a float for master stickiness

nsheridan:
This allows for < 1s stickiness, useful if you're using a database engine such as AWS Aurora which can routinely achieve lower than 1s replication delay.

### DIFF
--- a/lib/makara/cache.rb
+++ b/lib/makara/cache.rb
@@ -20,7 +20,7 @@ module Makara
       end
 
       def write(key, value, ttl)
-        store.try(:write, key, value, :expires_in => ttl.to_i)
+        store.try(:write, key, value, :expires_in => ttl.to_f)
       end
 
       protected
--- a/lib/makara/cache/memory_store.rb
+++ b/lib/makara/cache/memory_store.rb
@@ -13,14 +13,14 @@ module Makara
 
       def write(key, value, options = {})
         clean
-        @data[key] = [value, Time.now.to_i + (options[:expires_in] || 5).to_i]
+        @data[key] = [value, Time.now.to_f + (options[:expires_in] || 5).to_f]
         true
       end
 
       protected
 
       def clean
-        @data.delete_if{|k,v| v[1] <= Time.now.to_i }
+        @data.delete_if{|k,v| v[1] <= Time.now.to_f }
       end
 
     end


lib/makara/cache.rb
```diff
@@ -20,7 +20,7 @@ def read(key)
       end
 
       def write(key, value, ttl)
-        store.try(:write, key, value, :expires_in => ttl.to_i)
+        store.try(:write, key, value, :expires_in => ttl.to_f)
```
bf4, line 23: not all stores support floats, though memory does.

bleonard: what do those store do if they get a float? `to_i`?

bf4: e.g. https://github.com/sorentwo/readthis/issues/41

jeremy: Workaround: Store the stickiness duration as the cache value instead of implying it as the TTL. For cache TTL, use `ttl.ceil` so it's at least 1 second.

jeremy (changes requested):

mariokostelac:
It is quite compelling to use subsecond TTLs, but I do not think it's an easy one to implement unless you use a store natively supporting that (memcached does not as far as I can see). Wrapping an expiration into value is quite brittle because of clock drift (comparing clocks with high resolution [millis] on two different machines is just undefined behaviour).

We've used this patch and passing floats caused DalliStore rounding the TTL to 0 (because of https://github.com/petergoldstein/dalli/blob/f41bb9cf078246bd58a249cf09b1f56044e4662c/lib/dalli/client.rb#L277), which means "cache forever". You could argue that it is a deficiency of DalliStore (I could agree here with you), but even if that was not the case (if it was rounded to 1), stickiness would stay there 1 second, which makes this patch void. Unless we can think of widely used store supporting below second TTLs, I believe this patch is bringing just trouble.

bleonard:
I think subsecond might work when this gets merged in: https://github.com/taskrabbit/makara/pull/194

mariokostelac:
@bleonard that is going to work just for cookie storage, right? It'd be nice if other storages failed in that case.

rosa:
@mariokostelac, with #194 no other storages would be used for master stickiness, only cookies (it was motivated by #162). Setting a different storage would be deprecated after that.

mariokostelac:
Thanks for the context 😊

PR #150, opened 2017-03-02
Rails 5.1 compat: coerce Relation arg to SQL string before matching against it

### LINKED ISSUE #149: Incompatible with Rails 5.1 due to select_rows method signature change
Makara hijacks `select_rows` and inspects the first argument, expecting it to be a SQL string, to determine whether to route to primary or replica connection.

In https://github.com/rails/rails/pull/25522 the method signature changed to take an Arel relation instead of a SQL string, so all queries will go to the primary database and none to the replica.

To fix, we'll need to either expect & duly inspect the Relation or call `#to_sql` before checking. The first will take more code and possibly introduce bugs; the second will mean a perf hit due to converting every relation to SQL twice.

jeremy:
Makara hijacks `select_rows` and inspects the first argument, expecting
it to be a SQL string, to determine whether to route to primary or
replica connection.

In https://github.com/rails/rails/pull/25522 the method signature
changed to take an Arel relation instead of a SQL string, so all queries
will go to the primary database and none to the replica.

To fix, we'll need to either expect & duly inspect the Relation or call
`#to_sql` before checking. The first will take more code and possibly
introduce bugs; the second will mean a perf hit due to converting every
relation to SQL twice.

Fixes #149.

### DIFF
--- a/lib/active_record/connection_adapters/makara_abstract_adapter.rb
+++ b/lib/active_record/connection_adapters/makara_abstract_adapter.rb
@@ -165,24 +165,33 @@ module ActiveRecord
 
 
       def should_stick?(method_name, args)
-        sql = args.first.to_s
+        sql = coerce_query_to_sql_string(args.first)
         return false if sql_skip_stickiness_matchers.any?{|m| sql =~ m }
         super
       end
 
 
       def needed_by_all?(method_name, args)
-        sql = args.first.to_s
+        sql = coerce_query_to_sql_string(args.first)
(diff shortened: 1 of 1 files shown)


jeremy:
Test failure appears to be a fluke or other flakiness.

jeremy:
Note: We would've caught this immediately if Travis tested against Rails master (perhaps with failures allowed).

bleonard:
I'll merge this into master which is now checking against all the rails versions. 🎉

PR #152, opened 2017-04-13
Fix handling of blacklisting with frozen time.

jfilipe:
For frozen pre-epoch dates (< 1970), we were incorrectly assuming the
node was blacklisted since `Time.now.to_i` would be a negative number.
Instead, explicitly check if a `blacklisted_until` value was set before
doing a time range comparison.


bleonard:
@jfilipe thanks for the PR. I'm on vacation and finally getting around to makara management. Let me know how it's going for you.

PR #156, opened 2017-06-23
only check cache for "previously stuck to master" once per context

gapatrick:
Hi TaskRabbit people, I'd like to propose this change to help reduce usage of the context cache.

We use makara in a master-slave, multi-app server rails environment.  We use redis for centralized caching and for Rails.cache.  We noticed that over 90% of the commands processed by our redis master are for makara contexts.

Looking through makara code, and debugging, I found that in _appropriate_pool in proxy.rb, if the current context is NOT master, then every DB request during the processing of a single web request calls previously_stuck_to_master, which queries redis (in our case).  A lot of our web requests are page loads that do a lot (say, 40+) DB requests, so it also hits redis 40+ times.

However, during handling a web request, previously_stuck_to_master is either true or false as of the first use of _appropriate_pool, and it won't (and can't) change during the handling of that web request.  So it would be much more efficient to just check previously_stuck_to_master once.  If previously_stuck_to_master returns true, then the current context is switched to master, and any subsequent _appropriate_pool calls will use master because current == master (this is existing behaviour).

It's a bit convoluted to try to explain but hopefully that makes sense.

The fix I made just uses a thread local var to track whether the previous context has been checked yet or not.  It is set to false when Makara::Context.set_previous is called (it's called via middleware.rb before handling a web request), and then it is set to true in _appropriate_pool after previously_stuck_to_master is called.

If you want to change the method names or signatures I added, feel free, or let me know and I'll change them.

Thanks!
Greg Patrick - greg.patrick@varsitytutors.com

### DIFF
--- a/lib/makara/context.rb
+++ b/lib/makara/context.rb
@@ -13,15 +13,24 @@ module Makara
       end
 
       def get_previous
-        get_current_thread_local_for(:makara_context_previous)
+        get_current_thread_local_context_for(:makara_context_previous)
       end
 
       def set_previous(context)
+        set_current_thread_local(:makara_checked_previous,false)
         set_current_thread_local(:makara_context_previous,context)
       end
 
+      def checked_previous?
+        get_current_thread_local_for(:makara_checked_previous)
+      end
+
+      def set_checked_previous
+        set_current_thread_local(:makara_checked_previous,true)
+      end
+
       def get_current
-        get_current_thread_local_for(:makara_context_current)
+        get_current_thread_local_context_for(:makara_context_current)
       end
 
       def set_current(context)
@@ -32,7 +41,11 @@ module Makara
 
       def get_current_thread_local_for(type)
         t = Thread.current
-        current = t.respond_to?(:thread_variable_get) ? t.thread_variable_get(type) : t[type]
+        t.respond_to?(:thread_variable_get) ? t.thread_variable_get(type) : t[type]
+      end
+
+      def get_current_thread_local_context_for(type)
+        current = get_current_thread_local_for(type)
         current ||= set_current_thread_local(type,generate)
       end
 
--- a/lib/makara/context.rb
+++ b/lib/makara/context.rb
@@ -17,16 +17,8 @@ module Makara
       end
 
       def set_previous(context)
-        set_current_thread_local(:makara_checked_previous,false)
+        set_current_thread_local(:makara_cached_previous,nil)
         set_current_thread_local(:makara_context_previous,context)
-      end
-
-      def checked_previous?
-        get_current_thread_local_for(:makara_checked_previous)
-      end
-
-      def set_checked_previous
-        set_current_thread_local(:makara_checked_previous,true)
       end
 
       def get_current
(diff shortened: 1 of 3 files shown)


bleonard:
Thanks for the PR @vt-gpatrick 
What would you think about storing the boolean value of `!!Makara::Cache.read("makara::#{Makara::Context.get_previous}-#{@id}")` in the thread instead of if it checked?

Basically, we can memoize that `Cache.read` in `previously_stuck_to_master?` and keep calling `previously_stuck_to_master?` from above. I'm not fully clear on the nuances at play here and how they all get called especially with `@sticky` but that would seem to have the same effect with slightly less risk.

gapatrick:
Hi @bleonard, yes, that's a good idea.  I didn't think of that but it definitely makes sense.  I'll make some changes.  Thanks!

gapatrick:
Hi @bleonard I updated my PR with the changes you suggested.  I like it better now.  Added a couple more unit tests too.

bleonard:
Thanks! Looks great.

lib/makara/context.rb
```diff
@@ -13,26 +13,49 @@ def generate(seed = nil)
...
+      def cache_current(config_id, ttl)
+        Makara::Cache.write("makara::#{get_current}-#{config_id}", '1', ttl)
+      end
+
+      def cached_previous?(config_id)
+        cached_previous = get_current_thread_local_for(:makara_cached_previous)
+        # if haven't memoized the result of the Cache.read yet
+        if cached_previous.nil?
+          cached_previous = !!Makara::Cache.read("makara::#{Makara::Context.get_previous}-#{config_id}")
+          set_current_thread_local(:makara_cached_previous,cached_previous)
```
jeremy, line 45: Note that calling this with one `config_id` will set a thread-local cache that's used for all subsequent config_id, whether they were stuck to master or not.

lib/makara/proxy.rb
```diff
@@ -74,16 +74,16 @@ def hijacked?
...
       strategy_name = @config_parser.makara_config["#{role}_strategy".to_sym]
       case strategy_name
-      when 'round_robin', 'roundrobin', nil, ''
-        strategy_name = "::Makara::Strategies::RoundRobin"
-      when 'failover'
-        strategy_name = "::Makara::Strategies::PriorityFailover"
+        when 'round_robin', 'roundrobin', nil, ''
+          strategy_name = "::Makara::Strategies::RoundRobin"
+        when 'failover'
+          strategy_name = "::Makara::Strategies::PriorityFailover"
```
jeremy, line 86: Unrelated style change

lib/makara/proxy.rb
```diff
@@ -194,27 +194,26 @@ def _appropriate_pool(method_name, args)
...
+        # all slaves are down (or empty)
       elsif @slave_pool.completely_blacklisted?
         stick_to_master(method_name, args)
         @master_pool
 
       elsif in_transaction?
         @master_pool
 
-      # yay! use a slave
+        # yay! use a slave
```
jeremy, line 216: Ditto, unrelated style changes

lib/makara/context.rb
bleonard: Think we should have a hash to keep it bound to the `config_id`?

bleonard: ok, it looks like you did something like this in: https://github.com/taskrabbit/makara/pull/158 - will look there.

PR #158, opened 2017-06-29
Context: thread-local cache for previous context stickiness

jeremy:
* Fix that previous-stickiness check was cached for all configs
* Check Ruby Thread capabilities once rather than at runtime
* Choose thread-local get/set strategy once when we define the fetch/set methods rather than on every method call.
    
References ae2be27016b8866b6d9f86b6e1ad5699abd84f7a and #156

### DIFF
--- a/lib/makara/context.rb
+++ b/lib/makara/context.rb
@@ -13,55 +13,69 @@ module Makara
       end
 
       def get_previous
-        get_current_thread_local_context_for(:makara_context_previous)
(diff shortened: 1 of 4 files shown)


bleonard:
@jeremy I just merged in a fix from @vt-gpatrick that lowered the number of checks to the `Rails.cache` that conflicts with this one.
Could you port this over to master?

jeremy:
Rebased on master and updated #156 to support multiple configs.

PR #159, opened 2017-06-29
Configurable cookie options

jeremy:
Set the domain, mark as secure-only, SameSite, longer max age, etc.

### DIFF
--- a/lib/makara/middleware.rb
+++ b/lib/makara/middleware.rb
@@ -10,9 +10,15 @@ module Makara
 
     IDENTIFIER = '_mkra_ctxt'
 
+    DEFAULT_COOKIE = {
+      :path => '/',
+      :http_only => true,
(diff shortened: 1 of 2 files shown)


bleonard:
does your config have a longer max age?
What params are you using?
Maybe there are better defaults.

jeremy:
Setting `secure: true` and considering removing max-age (setting `max_age: nil`) so the cookie gets browser session lifetime. Further aim: keep stickiness context in the cookie instead of a backend cache, which itself needs to be replicated in a multi-datacenter environment. A stateless backend is much easier & cheaper to implement and operate.

bleonard:
Good idea. Revisiting this situation years later and I can't remember the reason for that store. I don't see why that wouldn't work.
I made this issue. https://github.com/taskrabbit/makara/issues/162

PR #160, opened 2017-06-29
Test against Rails 5.x and Ruby 2.x

jeremy:
* Rails 5.0 and 5.1
* Ruby 2.2, 2.3, 2.4, and 2.5-dev
* Dev: use default rbenv/chruby/rvm Ruby version
* Travis: cache bundle between builds


bleonard:
Thanks @jeremy ---- green!

PR #174, opened 2017-09-25
Fixes `can't add a new key into hash during iteration` error

ankane:
This adds thread-safety to the `MemoryStore`. Here's code to reproduce the current issue.

```ruby
Thread.abort_on_exception = true

$store = Makara::Cache::MemoryStore.new

workers = 10.times.map do
  Thread.new do
    loop do
      $store.write(rand(100), true, expires_in: 1)
      sleep(rand / 100)
    end
  end
end

workers.map(&:join)
```

### DIFF
--- a/lib/makara/cache/memory_store.rb
+++ b/lib/makara/cache/memory_store.rb
@@ -4,6 +4,7 @@ module Makara
 
       def initialize
         @data = {}
+        @mutex = Mutex.new
       end
 
       def read(key)
@@ -20,7 +21,9 @@ module Makara
       protected
 
       def clean
-        @data.delete_if{|k,v| v[1] <= Time.now.to_i }
+        @mutex.synchronize do
+          @data.delete_if{|k,v| v[1] <= Time.now.to_i }
+        end
       end
 
     end
--- a/spec/cache_spec.rb
+++ b/spec/cache_spec.rb
@@ -52,6 +52,28 @@ describe Makara::Cache do
       expect(data).not_to have_key('test')
     end
 
+    it 'has thread-safety' do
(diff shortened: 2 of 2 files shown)


bleonard:
Thanks @ankane - Think you could add the test case to the cache_spec?

ankane:
Hey @bleonard, thanks for the response. Added a test case.

bleonard:
@ankane - I've added you as a member if you'd like to participate with makara. Seems like a good idea if you are depending on it.

If so, want to test out master and tag a new release, etc?

ankane:
Yeah, happy to help out. I'll defer to you on when to release new versions. We'll try to get on master in the near future.

ankane:
Hey @bleonard, looks like I was invited to join org rather than this specific project.

bleonard:
It was a team that was only on makara projects. Now given project access if that's better for you.

ankane:
Ah, gotcha. All good now. Thanks.

bleonard:
Interesting that Ruby 2 is freaking out (no log) on Travis. https://travis-ci.org/taskrabbit/makara/builds/280541383

Some sort of Mutex or thread difference?

ankane:
That's odd. Can you try restarting the jobs?

bleonard:
Actually the nextval one (https://github.com/taskrabbit/makara/pull/173) failed first all in Ruby 2. Then this one, so likely not this one. Likely not that one either based on it's simplicity.

bleonard:
https://travis-ci.org/taskrabbit/makara/jobs/280541385 bundler failing to install in that ruby. tried similar setup locally and it worked.

ankane:
@bleonard fixed in #176

PR #179, opened 2017-10-30
add an option to skip process around context

yoonseokho:
This is a suggestion to make app more efficient for those who do not use context. 
You can skip processing around context by setting Makara::Context.skip_context_process.


bleonard:
Was it not helpful to you? What was the reason to consider skipping?

PR #184, opened 2017-11-29
Allow SELECTs that use common table expressions to go to replicas

ankane:
Formula is `WITH` followed by `) SELECT`. Most CTEs will have a SELECT somewhere in the query, but only if it's the last one should it be preceded by a right parenthesis (and optional whitespace).


palexvs:
CI is red after this change: https://travis-ci.org/taskrabbit/makara/builds/332157957

PR #187, opened 2018-01-03
Clarify README's "What goes where"

NobodysNightmare:
The original description implied that there was only one edge case, but there are more edge cases.
Thus after reading the README, I got the impression that makara would handle some cases incorrectly. However, after reading the code, I am now think that those cases are handled correctly.

### References

* [Makara handling transactions](https://github.com/taskrabbit/makara/blob/master/lib/makara/proxy.rb#L221-L222)
* [Makara handling locking reads](https://github.com/taskrabbit/makara/blob/master/lib/active_record/connection_adapters/makara_abstract_adapter.rb#L111)

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,7 +120,13 @@ By creating a makara database adapter which simply acts as a proxy we avoid any 
 
 ### What goes where?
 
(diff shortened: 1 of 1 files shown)


bleonard:
Thanks!

PR #191, opened 2018-01-26
Rollback not finished transactions

palexvs:
Makara gracefully handles exceptions, in the result, it does not rollback transactions on a master when the master goes into reboot/down. This PR should fix it.

### DIFF
--- a/lib/active_record/connection_adapters/makara_abstract_adapter.rb
+++ b/lib/active_record/connection_adapters/makara_abstract_adapter.rb
@@ -42,7 +42,7 @@ module ActiveRecord
           when *harsh_errors
             harshly(e)
           else
-            if connection_message?(e) || custom_error_message?(connection, e)
+            if !connection.in_transaction? && (connection_message?(e) || custom_error_message?(connection, e))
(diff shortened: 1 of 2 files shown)


bleonard:
Let's get a test case in the specs.

bleonard:
Thank you for the PR!

So now, if it is in a transaction and there is an issue, it will not failover. Maybe that's the right thing but it's worth understanding more.

It might be right because because a transaction makes everythig one atomic thing. And we won't be able to walk back and do them on the next server it falls back to.

One weird thing here is that it would not blacklist the connection (that happens in `gracefully`. So it might happen to the next one too. So there's something to thing about there.

palexvs:
> One weird thing here is that it would not blacklist the connection (that happens in gracefully. So it might happen to the next one too. So there's something to thing about there.

Good point. I will think too. Can you describe an architecture with a few masters? Currently, I have:
```
    connections:
      - role: master
        host: <%= SENV['MYSQL_HOST_MASTER'] || '127.0.0.1' %>
        disable_blacklist: true
```

PR #192, opened 2018-02-01
Fix loading fixtures in Rails 5.2

georgeclaghorn:
Preserve blocks when delegating `ConnectionWrapper` methods to `Proxy`. Rails’s newly-optimized fixture loading code calls `#transaction` from within a connection adapter method, exercising this code path. Without this fix, we get a `LocalJumpError` because `#transaction` expects a block.

/cc @jeremy @rosa


jeremy:
🎉

PR #193, opened 2018-02-12
Added secure flag by default if rack env shows protocol/scheme is https

amasses:
While it is possible to specify `secure: true` when using the middleware, its pretty difficult to set this, and I think that this should be a default, rather then a difficult opt-in.

### DIFF
--- a/lib/makara/middleware.rb
+++ b/lib/makara/middleware.rb
@@ -30,7 +30,7 @@ module Makara
       Makara::Context.set_current new_context(env)
 
       status, headers, body = @app.call(env)
-
+      @cookie[:secure] = is_https_request(env) unless @cookie.keys.include?(:secure)
       store_context(status, headers)
 
       [status, headers, body]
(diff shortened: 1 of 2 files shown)


bleonard:
Looks good to me. Is there a use case where someone wouldn't want us to mess with this?

amasses:
Could be where someone is running across both HTTPS and HTTP, however I think this use case is getting much smaller.

I suspect if this is the case they should be able to load the middleware themselves and specify secure: false (though the reason for this PR was because doing that was overly complicated...

eugeneius:
While mixing HTTP and HTTPS is less common than it used to be, it seems like overreach for us to set a secure cookie whenever the current request is over HTTPS.

For Rails apps, we could default to secure cookies if `config.force_ssl` is true, i.e. change:
https://github.com/taskrabbit/makara/blob/560dc333f76138c09f25f1fd3afa2d3da473f6a2/lib/makara/railtie.rb#L5

To:
```ruby
  app.middleware.use Makara::Middleware, secure: app.config.force_ssl
```

Non-Rails apps would still have to manually pass the `secure` option - but they already have to add the middleware manually, so it's relatively easy to configure.

ankane:
fwiw, `config.force_ssl` already makes all cookies secure.

https://github.com/rails/rails/blob/3d70e9609760cb3046cf219cd49f3e9a838a0291/actionpack/lib/action_dispatch/middleware/ssl.rb#L20

Issue #162, opened 2017-07-05
Only use cookies for context


bleonard:
Frankly, I don't fully remember why we had the context store in `Rails.cache` - There are random uids and they point to data in Redis. We store the uid in the cookie and reference the data. But that data is only a "1" meaning it should stick to master. I assume we had bigger plans for it have more info, but it never happened.

Given that's all it is, it would be smart to be able to encode the "stick to master" bit in the Rails session or it's own cookie. Maybe a hash in case we need some other bit. This would remove the need for the backend store and be generally more manageable.

johnwu96822:
@bleonard What about API calls that do not use cookies?? https://github.com/taskrabbit/makara/issues/175

bleonard:
The old method requires cookies, too. It has to store the "session context" guid somewhere. One way or another, that contet has to be saved. In our iOS app (using AFNetworking), for example, there is a way to actually send the right cookies headers back and forth.

rosa:
Hey @bleonard! This week I've been working on a `CookieStore` for Makara, since right now we are using it without stickiness because we don't set any kind of backend cache (we just use the `NoopStore`) and wanted to use cookies for this. While doing that, I realised that I didn't quite understand what the real intended behaviour for stickiness was, as it's implemented now. 

Without looking much as the code, I understood that once you stick to `master`, you want that _all_ operations happening within the set `master_ttl` for that config go to `master` as well. That's why you set that ttl larger than the replication lag you can tolerate. However, once I was deep in the code, what I saw is that only the first request after sticking (or the first two requests if the first one is a redirect) goes to the `master` as well. Let me explain:

These are the relevant lines, when deciding whether to use the master, and whether to stick or not for the next request https://github.com/taskrabbit/makara/blob/e95d4303d350c9ab9dfc41b1e5c2743a1849b9c4/lib/makara/proxy.rb#L201-L214

It's easier to see with an example. Suppose we start without a context, that is, without a cookie set or anything, and make 3 requests that take less than the TTL we have configured for sticking to the master. In each request we keep track of the previous context and the current context, and after the request we set the cookie with the current context. 

Request 1, this one needs the master because it writes:
```
cookies = {}
previous_context = "P1" # Generated, since the cookie is not set
current_context = "C1" # Generated. If the previous request was a redirect, we'd set it to P1.
cache = {}

appropriate_pool = primary # Since it's a write operation
# Stick to master
cache["C1"] = true
# Response cookies: _mkra_ctxt=C1--201
```

[...]

PR #194, opened 2018-02-12
Implement stickiness for the duration of `master_ttl` via cookies

rosa:
This pull request significantly changes how Makara's context and stickiness behaves. It implements what is described in [this comment](https://github.com/taskrabbit/makara/issues/162#issuecomment-363008007), and can be summarised as:
* Stickiness is not maintained only for the subsequent request after using `master`, is maintained for the whole duration of `master_ttl`, as many as request that be. I noticed [another old comment](https://github.com/taskrabbit/makara/pull/100#issuecomment-163882651) in which this behaviour was assumed by another person, as a more intuitive way of mitigating replication lag.
* Makara no longer keeps track of any context identifier or hash. It only keeps track of which configurations are stuck to `master` and when that stickiness expires. This is what is considered _context_ now: a map of configuration ids and their stickiness expiration times.
* Context is maintained between requests stored in a cookie. We don't store a hash anymore, we store the whole context (the map of config ids and timestamps). 
* It's still possible to skip stickiness, force stickiness, and release all stuck connections at once.

Todo:
- [ ] Support passing stickiness information via query args, as it's currently possible with `http://mysite.com/path/to/resource?_mkra_ctxt=#{context}`.

----

This is a very big change in code and in behaviour, so I totally understand any reluctance to merge it 😅 If this is too radical, I'm happy to get to an intermediate solution, although I do believe that the change is behaviour makes more sense as the idea of replication lag mitigation goes.


bleonard:
:boom: Thanks @rosa 
I'll look at it this week.
Have you put this into use yet in your app?

rosa:
@bleonard not yet! I have only tested it locally, but I'll let you know if we end up using it in production, I'll work on that this week 🤞

README.md
```diff
@@ -63,21 +63,16 @@ Makara comes with a config parser which will handle providing subconfigs to the
 
 ### Context
 
-Makara handles stickyness by keeping track of a context (sha). In a multi-instance environment it persists a context in a cache. If Rails is present it will automatically use Rails.cache. You can provide any kind of store as long as it responds to the methods required in [lib/makara/cache.rb](lib/makara/cache.rb).
+Makara handles stickiness by keeping track of which configurations are stuck at any given moment. The context is basically a hash of configuration ids and the timestamp until which they are stuck.
 
-```ruby
-Makara::Cache.store = MyRedisCacheStore.new
```
bleonard: It might be useful to still support this `store=` setter and give a deprecation warning.

README.md
```diff
@@ -63,21 +63,16 @@ Makara comes with a config parser which will handle providing subconfigs to the
...
-```
+To handle persistence of context across requests in a Rack app, makara provides a middleware. It lays a cookie named `_mkra_ctxt` which contains the current context. If the next request is executed before the cookie expires, that given context will be used. If something occurs which naturally requires master on the second request, the context is updated and stored again.
```
bleonard: You changed to cookie name to `_mkra_stck`

README.md
```diff
@@ -63,21 +63,16 @@ Makara comes with a config parser which will handle providing subconfigs to the
...
 
-#### Changing Context
-
-If you need to change the makara context, releasing any stuck connections, all you have to do is:
+If you need to clear the current context, releasing any stuck connections, all you have to do is:
 
 ```ruby
-ctx = Makara::Context.generate # or any unique sha
-Makara::Context.set_current ctx
+Makara::Context.release_all
```
bleonard: Should they be able to release a single one?

README.md
```diff
@@ -275,16 +270,7 @@ def handle_request_after_third_party_record_creation
...
-```ruby
-def send_url_to_third_party
-  context = Makara::Context.get_current
-  ThirdParty.read_from_here!("http://mysite.com/path/to/resource?_mkra_ctxt=#{context}")
-end
-```
-
 ## Todo
 
-* Cookie based cache store?
```
bleonard, line 289: 💥

lib/makara/context.rb
```diff
@@ -1,83 +1,144 @@
...
+# Keeps track of the current stickiness state for different Makara configurations
 module Makara
   class Context
-    class << self
 
-      def generate(seed = nil)
-        seed ||= "#{Time.now.to_i}#{Thread.current.object_id}#{rand(99999)}"
-        Digest::MD5.hexdigest(seed)
-      end
+    IDENTIFIER = '_mkra_stck'
```
bleonard: in our app, we usually `.freeze` these. it's a habit now. have you seen benefits there?

lib/makara/context.rb
```diff
@@ -1,83 +1,144 @@
...
 
-      def get_previous
-        fetch(:makara_context_previous) { generate }
-      end
+    DEFAULT_OPTIONS = {
+      path: "/",
+      http_only: true
```
bleonard: There's another PR for this: https://github.com/taskrabbit/makara/pull/193
that makes it `secure` if applicable. Seems like a good default. We can do it later, but noting here in case relevant to the new design.

lib/makara/context.rb
```diff
@@ -1,83 +1,144 @@
 require 'digest/md5'
 
-# Keeps track of the current and previous context (hexdigests)
-# If a new context is needed it can be generated via Makara::Context.generate
-
+# Keeps track of the current stickiness state for different Makara configurations
 module Makara
```
bleonard: Overall, this class has changed to being _about_ the context to also being about the cookies storage.
Before, the middle ware was the one _using_ the context and writing the cookies.
Do you believe there is there any value in keeping that separation?

I suppose that would be of value if there was to be another way to persist the context similar to how we were thinking before. I'm 50/50 here.

README.md
bleonard: I see that you did. Just not in the readme. That's fine. Nice.

lib/makara/context.rb
bleonard: It would move or less be:
1) the middleware knows the cookie name
2) `init` it with the current hash instead of the request.
3) moving the commit method the middleware. or maybe asking this object what to commit. It's hard to say if this class should know how to serialize (with the `|`, etc) or the middleware in that case.

lib/makara/context.rb
```diff
@@ -1,83 +1,144 @@
...
+      end
+
+      value = data.collect { |config_id, ttl| "#{config_id}:#{ttl}" }.join('|')
+      { :max_age => max_age, :value => value }
+    end
+
+    private
+
+    def expired?(timestamp)
+      timestamp <= Time.now.to_f
```
bleonard: It's kind of weird thing, but we'd want this to be the same across the entire request. With this code, there is a chance it would change halfway through and might produce strange results. I would suggest to save the time as an instance variable in the constructor.

lib/makara/context.rb
```diff
@@ -1,83 +1,144 @@
...
+
+    attr_accessor :data
+
+    def initialize(data)
+      @data = data
+      @dirty = false
+    end
+
+    def stick(config_id, ttl)
+      data[config_id] = Time.now.to_f + ttl.to_f
```
bleonard: let's say we cache time in the constructor as suggested below. This would still be `now`.
There's something interesting around saving the `ttl` and then making all the times at `commit` time, but that might add too much complexity.

lib/makara/context.rb
```diff
@@ -1,83 +1,144 @@
...
+
+    def release_all
+      if self.data.any?
+        self.data = {}
+        @dirty = true
+      end
+    end
+
+    def to_cookie_options
+      max_age = if data.any?
```
bleonard: there's something in the back of my mind about us being more dependent on the client's clock all of a sudden and that time coordinating with our server. Maybe that's not completely new (since the was a defaulted max_age before), but it might be worth thinking about. We are already resistant to the dates being stale because the value is in the cookie. Should we add some buffers for this max age?

lib/makara/context.rb
```diff
@@ -1,83 +1,144 @@
...
+        (data.values.max - Time.now.to_f).ceil + 1
+      else
+        0
+      end
+
+      value = data.collect { |config_id, ttl| "#{config_id}:#{ttl}" }.join('|')
+      { :max_age => max_age, :value => value }
```
bleonard: Is there `expires_at` and is that better or worse than `max_age`?

lib/makara/context.rb
```diff
@@ -1,83 +1,144 @@
...
+      private
+
+      # Pairs of {config_id}:{timestamp}, separated by "|"
+      # config_id1:1518270031.3132212|config_id2:1518270030.313232 ..
+      def parse(cookie_string)
+        return {} if cookie_string.empty?
+
+        states = cookie_string.split("|")
+        states.each_with_object({}) do |state, data|
+          config_id, timestamp = state.split(":")
```
bleonard: wherever we read in these config ids, we should consider making sure they are all simple strings. At the very least, they shouldn't have `|` or `:` - maybe `gsub` them out and output a warning?

lib/makara/context.rb
```diff
@@ -1,83 +1,144 @@
...
+    # to be persisted when the request finishes
+    def dirty?
+      @dirty
+    end
+
+    def release(config_id)
+      @dirty ||= !!data.delete(config_id)
+    end
+
+    def release_expired
```
bleonard: nowhere in here are we dealing with unknown `config_id`s. Think that's ok and they'll just age out?

lib/makara/context.rb
```diff
@@ -1,83 +1,144 @@
...
+    end
+
+    attr_accessor :data
+
+    def initialize(data)
+      @data = data
+      @dirty = false
+    end
+
+    def stick(config_id, ttl)
```
bleonard: I think people will be very happy if this ttl can be a float (< 1 second now) such as https://github.com/taskrabbit/makara/pull/143
Is there a test to make sure that works?

lib/makara/pool.rb
```diff
@@ -137,16 +135,7 @@ def connection_made?
...
-        con = @strategy.current
-        return con if con
-      end
-
-      con = @strategy.next
-      if con
-        @context = Makara::Context.get_current
-      end
-      con
+      (@strategy.current if @proxy.sticky) || @strategy.next
```
bleonard: I'm sure this is right, but it's more nuanced then we tend to do things. If I had to read it twice to know what it was doing, then it might cause a problem. Code is written once and debugged 10 times and read 1000s of times, so we optimize for the reading.

Maybe something like this?
```ruby
strategy = nil
strategy ||= @strategy.current if @proxy.sticky
strategy ||= @strategy.next
strategy
```

or

```ruby
return @strategy.current if @proxy.sticky && @strategy.current
@strategy.next
```

I'm not clear if `@strategy.current` can actually be `nil`. can it?

lib/makara/context.rb
bleonard: I guess in `ConfigParser`

lib/makara/proxy.rb
```diff
@@ -248,24 +240,22 @@ def hijacked
     end
 
 
-    def previously_stuck_to_master?
-      @sticky && Makara::Context.previously_stuck?(@id)
+    def stuck_to_master?
+      @sticky && !@skip_sticking && Makara::Context.stuck?(@id)
```
bleonard: Should this use `should_stick?` somehow? Or DRY that up?
It's just the two variables but seems semantically correct.
`should_stick?("stuck_to_master_check", []) && Makara::Context.stuck?(@id)`

lib/makara/proxy.rb
```diff
@@ -248,24 +240,22 @@ def hijacked
...
-      stick_to_master!(write_to_cache)
+      stick_to_master!
     end
 
 
-    # if we are configured to be sticky and we aren't bypassing stickiness
+    # If we are configured to be sticky and we aren't bypassing stickiness,
+    # (method and args don't matter)
     def should_stick?(method_name, args)
       @sticky && !@skip_sticking
```
bleonard: I wonder why we sent this method name and args here. probably so someone more clever than me could override them in some way. Not sure if that's needed.

lib/makara/proxy.rb
```diff
@@ -248,24 +240,22 @@ def hijacked
...
 
 
-    def previously_stuck_to_master?
-      @sticky && Makara::Context.previously_stuck?(@id)
+    def stuck_to_master?
+      @sticky && !@skip_sticking && Makara::Context.stuck?(@id)
     end
 
 
-    def stick_to_master(method_name, args, write_to_cache = true)
```
bleonard: As noted above, let's keep `write_to_cache`.

lib/makara/proxy.rb
```diff
@@ -59,22 +59,18 @@ def initialize(config)
...
-      @master_context ||= before_context
     end
 
     def hijacked?
       @hijacked
     end
 
-    def stick_to_master!(write_to_cache = true)
-      @master_context = Makara::Context.get_current
-      Makara::Context.stick(@master_context, @id, @ttl) if write_to_cache
```
bleonard, line 77: This `write_to_cache` thing seems important, but it really would complicate your `Context` checks.
Now that I think about it, things get a little weird in the new model with really long requests. It seems against the expected behavior to `stick` in a request, it be a long request, and then unstick later in that request. Let's call this "Full Request Sticking" and I'll put a note above in the context.

Here, let's keep the `write_to_cache` variable and pass it to the `Context`.

lib/makara/context.rb
bleonard: I'm rethinking this based on the `write_to_cache` and "Full Request Sticking" comments below.
What is this method was: `def stick(config_id, ttl, write_to_cache=true)`

The "Full Request Sticking" concept says that i should stick the whole request even if the request is longer than the `ttl`. So we need to set something. Maybe `data` should be a hash like:
`data[config_id] = { expires_at: time_from_cookie, ttl: ttl_from_this_method, persist: true }`

`initialize` would set all the `expires_at` and this method would set the `ttl` if it's bigger than current (or unset) and set `persist` to `true` if `write_to_cache` was `true`.

Then below, we'd use this data to write the cookies. I think this `Time.now` stuff is dangerous because of Transactions in long requests. The TTL shouldn't really "kick in" until the request is over because that's when the data actually starts to replicate. So in `to_cookie_options` we would get the `Time.now` and then add all the `ttl` values if `persist` was true.

lib/makara/context.rb
```diff
@@ -1,83 +1,144 @@
...
+      data[config_id] = Time.now.to_f + ttl.to_f
+      @dirty = true
+    end
+
+    def stuck?(config_id)
+      data[config_id] && !expired?(data[config_id])
```
bleonard: For "Full Request Sticking" concept, this method would look more like
```ruby
return false unless data[config_id]
return true if data[config_id][:ttl] # because it was asked to stick in this request
expired?(data[config_id][:expires_at])
```

lib/makara/context.rb
bleonard: In theory, you could also have a separate hash if that feels better and leave `data` as just what's in the cookie. Not sure if that is better or not. Seems like more to coordinate, but maybe there are advantages, too, when really going through it.

bleonard (commented):
This is great @rosa ! I tried to be thorough to show respect to the great work. Very exciting!

The main thing is "Full Request Sticking" which I mention a few times - that we should stick the whole request if it's a long one. Right now, I think the `Time.now` stuff might cause subtle issues in that world.

lib/makara/context.rb
bleonard: Just thinking this through one more time as I walked away. Usually when I make one of these hashes with several keys that have various meanings (often called a data clump), I don't regret making an Object or Struct to represent it. Then you can ask it these semantic questions. Might be worth thinking about.

README.md
rosa: Ooops, yes I did! I renamed it later on, thinking that using the same name could cause issues with already set cookies, and forgot to change this reference.

README.md
rosa: True! I need to add it to the readme 😅

rosa:
@bleonard thanks so much for your thorough review, I really appreciate it! I'll address each comment separately and will push changes soon. 

You made a very good point about long requests, I didn't think about that. What you propose with "full request sticking" makes sense to me. Unsticking in the middle of the request doesn't seem very problematic to me (if the `master_ttl` was chosen to allow for replication lag, queries that go to the replicas after that ttl has passed even if they happen in the same request should be fine in theory) but sticking way before the request is over does seem problematic! I mean the part in which sticking a config does `Time.now.to_f + ttl`, I definitely need to change that, and will explore implementing the whole full request sticking concept. 

Thanks again! 🙇‍♀️

README.md
rosa: Good idea 👍

lib/makara/context.rb
rosa: Good point. Maybe having a new class, `Makara::CookieParser` (or `Makara::CookieSerializer`, or `Makara::ContextCookie`... or just `Makara::Cookie`, I think that'd work), that knows how to build the hash from the cookie string and how to serialize the hash), and the whole thing would be:
- The middleware knows the cookie name, gets the contents from the request, uses that new class to build the hash, and passes that to `init`.
- The `commit` method is moved to the middleware, that gets what should be committed from `Context`, uses that new class to build the cookie string, and then modifies the headers.

I'll try that to see how it looks!

rosa: Hmm... although, thinking about it again, if the new class is called `Makara::Cookie`, it would imply it is who knows about the cookie name and it would be the one initialising the context and _committing_ the new cookie. If we want to keep that in the middleware, perhaps `Makara::ContextParser` or `Makara::ContextSerializer` or something like that would be better, for parsing the cookie contents, and for serialising them.

lib/makara/context.rb
rosa: Good question! The current implementation uses `max-age` and I thought whether it should be changed to `expires`. From what I read, it seemed that `expires` was deprecated on HTTP version 1.1, replaced by `max-age`. I understood that this was simpler and better because it isn't affected by clock differences between the server and the client browser. However, older versions of IE don't support `max-age`, so `expires` is still used in many places to ensure browser support. 

I just realised that when setting both, `max-age` takes precedence, so perhaps I should just set both 😄

lib/makara/context.rb
rosa: I added 1 second buffer (at least, since it uses `ceil`), do you think it should be bigger?

lib/makara/context.rb
rosa: Ahh, good point! I think this should work fine, but I haven't written any tests to show it. I'll do that.

lib/makara/context.rb
rosa: Ah, yes! Agreed, `ConfigParser` would probably be the best place to do the `gsub` and to show any warnings.

lib/makara/context.rb
rosa: Hmm... do you mean `config_id`s that don't correspond to any existing configuration? That is, if a configuration is changed for example, so the id changes? 
Yes, I thought in that case they would only live in the cookie until their expiration time passes, which should be very short in general. They'll be removed by this method when that happens.

bleonard: Yeah, that's what I meant. If someone were to stop using Makara for redis, would it be in the cookie forever. As it is now, it wold go away when the TTL is over, so that's seems fine as these should be fairly short. That small risk seems better than accidentally removing it in some transition situation. So carry on.

lib/makara/context.rb
bleonard: It might be nice to note that giving it an `id` in the yml is best practice to preserve stickiness even when a new db is added.

lib/makara/context.rb
bleonard: I don't see how 5 seconds could hurt things.

lib/makara/context.rb
bleonard: Whatever you think as you've looked into it more.

bleonard:
Great job @rosa, look forward to seeing the updates.

lib/makara/proxy.rb
rosa: Ahh! I forgot to comment on this! It's useful when implementing other kind of proxy, like for the AR adapter: https://github.com/taskrabbit/makara/blob/c6b158e847468379da8cc2eb69aee455bef386fb/lib/active_record/connection_adapters/makara_abstract_adapter.rb#L167-L171

bleonard: Oh yes, that's pretty key. I wonder if the above should use it with fake arguments or DRY up that check or leave it as-is. Unclear to me.

lib/makara/context.rb
rosa: True! The `README` could be a good place for that 🤔Around the place [where `database.yml` is explained](https://github.com/taskrabbit/makara#databaseyml). I should also add a different example, for Redis or something, to show how a different configuration that is not `database.yml` could look.

lib/makara/pool.rb
rosa: Ooh, yes! I think it can, if the current connection is blacklisted, for example. In that case this doesn't behave exactly as before, great catch!

rosa: Ah, no, no, it does behave as before. If that happens `@strategy.next` is used. But I agree this way is less clear, I'll change it to your second suggestion 👍

lib/makara/context.rb
rosa: I'll tackle this separately 👍

rosa:
Ok! This took me a bit longer than I thought 😅I think I have addressed now all the issues brought up. I have also done a fair bit of refactoring in my previous code. 

The main changes with respect to the original implementation are:

#### _Proxy_ instead of _Configuration_
Stop using the concept of _configuration id_, or _sticking a configuration_ because it's confusing and doesn't make a lot of sense. I started using this terminology because in `Makara::Context` we were passing `config_id` around, to build the cache key, so I continued using that. It made sense to me because I knew what I was talking about, but probably from someone not familiar with Makara this couldn't make sense (what does "sticking a configuration" can possibly mean?). Then I realised that the `id` defined in the configuration actually identifies the _proxy_, and in fact we use it as `@id` inside the `Proxy`. Changing this everything becomes much clearer. It's the proxy the one that gets stuck to `master`, not the configuration. 

#### _Full request sticking_
Make TTL to "kick in" at the end of the request, when we move on to the next context and persist it. This is the "full request sticking" we talked about: if we stick to `master` in a long request, it could happen that the stickiness expires within the same request. This is no longer the case. After thinking about how to do this for a while, I finally did this by having two different sets of data in the current context: _stored_ data and _staged_ data. The stored data comes from the cookie, it has already been stored before. Staged data is new for the request, and will be committed/stored at the end of the request. 

#### Sticking without persisting
Support `Proxy#stick_to_master!(persist = false)`. While doing the "full request sticking" part, I realised I hadn't taken this into account. Someone can decide they want to stick to master for the current request, for a specific operation, but without necessarily continuing being stuck for the duration of `master_ttl`. This was achieved before by `write_to_cache = false` and an instance variable `@master_context` in the proxy. If `write_to_cache` was `false`, the context was only stored in that instance variable. Then on subsequent queries, if the current context was `@master_context`, it means we are stuck, but we haven't persisted this to the cache for the next request at all. Doing the same kind of implementation in this case doesn't quite work because `Context.release[_all]` doesn't know about individual proxy instances and their instance variables. It worked previously because all we did was to change the current context identifier, which indirectly _expired" `@master_context`. However, I found that using the new stored and staged _areas_ in `Context`, I could implement this in an intuitive way: sticking without persisting is the same as adding a proxy to the staging area with a `ttl = 0`. That means: this proxy should be stick for the current request and no more. Then, `Context.release_[all]` obviously clears the stored and staged data, so we can expire non-persisted stickiness as well.

---

Apart from this I implemented the other minor issues you noticed, like showing a warning for `Makara::Cache.store = ..`, or gsubbing `|` and `:` in the proxy id. 

I still need to test this with our own app, and see if we can get it to production 🤞

README.md
```diff
@@ -61,40 +61,44 @@ This implementation will send any request not like "SELECT..." to a master conne
 
 Makara comes with a config parser which will handle providing subconfigs to the `connection_for` method. Check out the ActiveRecord database.yml example below for more info.
 
-### Context
+### Stickiness Context
 
-Makara handles stickyness by keeping track of a context (sha). In a multi-instance environment it persists a context in a cache. If Rails is present it will automatically use Rails.cache. You can provide any kind of store as long as it responds to the methods required in [lib/makara/cache.rb](lib/makara/cache.rb).
+Makara handles stickiness by keeping track of which proxies are stuck at any given moment. The context is basically a hash of proxy ids and the timestamp until which they are stuck.
```
bleonard: this sounds like we create a `hash` (like MD5 sort of thing) of those things. Maybe "The context keeps track of the expiration timestamp for each proxy id."

README.md
```diff
@@ -61,40 +61,44 @@ This implementation will send any request not like "SELECT..." to a master conne
...
 If you need to force master in your app then you can simply invoke stick_to_master! on your connection:
 
 ```ruby
-write_to_cache = true # or false
-proxy.stick_to_master!(write_to_cache)
+persist = true # or false, it's true by default
+proxy.stick_to_master!(persist)
 ```
 
+It'll keep the proxy stick to master for the current request, and if `persist = true` (default), it'll be also stored in the context for subsequent requests, keeping the proxy stuck up to the duration of `master_ttl` configured for the proxy.
```
bleonard: "stuck"

README.md
```diff
@@ -275,16 +282,7 @@ def handle_request_after_third_party_record_creation
...
-def send_url_to_third_party
-  context = Makara::Context.get_current
-  ThirdParty.read_from_here!("http://mysite.com/path/to/resource?_mkra_ctxt=#{context}")
-end
-```
-
 ## Todo
 
-* Cookie based cache store?
+* Support for providing context as query param
```
bleonard, line 287: Good idea!

lib/makara/context.rb
bleonard: :+1:

lib/makara/context.rb
```diff
@@ -1,83 +1,159 @@
...
+    attr_accessor :stored_data, :staged_data
+
+    def initialize(context_data)
+      @stored_data = context_data
+      @staged_data = {}
+      @dirty = @was_dirty = false
+    end
+
+    def stage(proxy_id, ttl)
+      staged_data[proxy_id] = ttl.to_f
```
bleonard: I think you'll want to do a `max` here of the `ttl` - imagine it did a `write_to_context=true` and then `write_to_context=false` in the same request. This would end up at zero but probably should have ended up persisting.

Maybe:
```ruby
[staged_data[proxy_id].to_f, ttl.to_f].max
```

lib/makara/context.rb
```diff
@@ -1,83 +1,159 @@
...
+    end
 
-      def get_previous
-        fetch(:makara_context_previous) { generate }
+    def was_dirty?
+      @was_dirty
+    end
+
+    def expired?(timestamp)
+      timestamp <= Time.now.to_f
```
bleonard: I still think we should cache the time (`@now = Time.now.to_f`) on initialization of this object so this is consistent for the entire length of the request. At least for use in the `stuck?` call above.

In `release_expired` below, it should be `Time.now` when it is calling release so this method might not be needed and just use the comparison in both of those spots to either the instance variable or `time = Time.now` at the top of `release_expired`.

lib/makara/context.rb
```diff
@@ -1,83 +1,159 @@
...
+    def release_expired
+      previous_size = stored_data.size
+      stored_data.delete_if { |_, timestamp| expired?(timestamp) }
+      @dirty ||= previous_size != stored_data.size
+    end
+
+    def store_staged_data
+      staged_data.each do |proxy_id, ttl|
+        if ttl > 0
+          self.stored_data[proxy_id] = Time.now.to_f + ttl
```
bleonard: Maybe we need a `max` too? If the previous request actually had a longer timestamp, we'd want to pass that through here.

lib/makara/context.rb
```diff
@@ -1,83 +1,159 @@
...
       end
+      self.staged_data = {}
+    end
+
+    # Stores the staged data with an expiration time based on the current time,
+    # and clears any expired entries. Returns true if any changes were made to
+    # the current store
+    def commit
+      release_expired
+      store_staged_data
```
bleonard: do we want to `store_staged_data` before the `release_expired` so that the `0` values don't get written?

lib/makara/context.rb
```diff
@@ -1,83 +1,159 @@
...
+        def set(key, value)
           Thread.current[key]=value
         end
       end
 
+      private
+
+      # Pairs of {proxy_id}:{timestamp}, separated by "|"
+      # proxy_id1:1518270031.3132212|proxy_id2:1518270030.313232 ..
+      def parse(cookie_string)
```
bleonard: I don't think this is used anymore (or there is a duplicate in the new `Cookie` class).

lib/makara/cookie.rb
```diff
@@ -0,0 +1,50 @@
...
+      path: "/",
+      http_only: true
+    }
+
+    def fetch(request)
+      parse(request.cookies[IDENTIFIER].to_s)
+    end
+
+    def store(context_data, headers, options = {})
+      unless context_data.nil?
```
bleonard, line 19: If I follow everything correctly, this could be `nil` if there were no changes to the cookie (`@was_dirty == false`) - I'm not clear on the situation. Do you have to set the header here to keep the cookie around? That is it was a ttl what was longer than the request so there were no changes but there still needs to be a cookie. In other words, does it pass through by default? In which case, we do we delete the cookie in some case?

lib/makara/cookie.rb
```diff
@@ -0,0 +1,50 @@
...
+        context_data[proxy_id] = timestamp.to_f if proxy_id && timestamp
+      end
+    end
+
+    def build_cookie(context_data, options)
+      cookie = DEFAULT_OPTIONS.merge(options)
+      now = Time.now
+
+      cookie[:max_age] = if context_data.any?
+        (context_data.values.max - now.to_f).ceil + 5
```
bleonard: might be nice to make this `5` a `BUFFER_TIME` constant or something to be clear what it is semantically.

lib/makara/proxy.rb
```diff
@@ -248,28 +243,28 @@ def hijacked
...
 
 
-    def previously_stuck_to_master?
-      @sticky && Makara::Context.previously_stuck?(@id)
+    def stuck_to_master?
+      sticky? && Makara::Context.stuck?(@id)
     end
 
-
-    def stick_to_master(method_name, args, write_to_cache = true)
```
bleonard, line 255: do we need to pass through that `write_to_cache=true` (now `persist=true`) thing here to `stick_to_master!` ?

README.md
bleonard: Maybe `path/here?makara_stick=proxy_one|proxy_two`
Not sure if need timestamps. In that case could just be the whole cookie string.

bleonard (commented):
Looking great @rosa!
Just a few notes on possible edge cases and cookie lifecycle.

README.md
rosa: You are right, I was influenced by the `Hash` class name, perhaps naming it a _dictionary_ or a _mapping_ is better.

README.md
rosa: 😅 Thanks!

README.md
rosa: Yes, something like that! Either the whole cookie string, or just one, and merge it together with the cookie string 🤔

lib/makara/context.rb
rosa: Oohh, good point! I'll add a test for this! Not only in the same request, but also if it came from the previous request and it still hasn't expired.

lib/makara/context.rb
rosa: Got it! I did this on purpose, I wasn't sure how you preferred to handle proxies that had been stuck in the previous request. In the current implementation with cache, if the cache expires in the middle of the request and you haven't checked whether a proxy is stuck before (and so caching the `@master_context` in the proxy object), the proxy won't be stuck. I kept that same behaviour here, but I think it's totally reasonable to say: a proxy will be stuck if the request starts before its current `expires_at` timestamp, even if it expires when the request is not over.

lib/makara/context.rb
rosa: Oh, yes! That's a really edge case (since in general `ttl` will be always the configured `master_ttl`), but it can totally happen. I'll add also a test for this.

lib/makara/context.rb
rosa: Hmm... the `0` values won't get written in any case, since we'll only store `proxy_id, ttl` pairs with `ttl > 0`. I don't see any other reason why we should check for expiration of data staged in this request, in theory no staged data should be expired because we'll be comparing `Time.now + ttl` (with `ttl > 0`) with `Time.now`, so I believe this is the right order. Am I missing anything?

lib/makara/context.rb
rosa: Oooops, yes! I moved it to `Cookie` and forgot to remove it from here! 😅

lib/makara/cookie.rb
rosa: > this could be nil if there were no changes to the cookie 

Correct!

> Do you have to set the header here to keep the cookie around?

No! If you don't send anything the cookie will be kept as it was. That was my idea of returning `nil` here: there is no _next_ context => then there is no cookie sent.

> In which case, we do we delete the cookie in some case?

We do! When the _next_ context exists and it's empty, we send a cookie with `max-age=0` and `expires=<current time>`. That's a way of _deleting_ the cookie.

lib/makara/cookie.rb
rosa: Yes, much better, great suggestion.

lib/makara/proxy.rb
rosa: I think it's not necessary here, because in all cases we call `stick_to_master`, which is private to the proxy, it's because we came across some action that made us to use `master`, and in that case we always want to persist the value for the `master_ttl` duration.

rosa:
Thanks for the second thorough review @bleonard! I'm quite enjoying working on this 😄

rosa:
Last set of changes pushed. Last week at the end I couldn't get to try this with our app, I'll see if I can do that this week 💪

lib/makara/config_parser.rb
```diff
@@ -197,5 +197,15 @@ def recursive_sort(thing)
 
     end
 
+
+    def sanitize_id(id)
+      return if id.nil? || id.empty?
+
+      id.gsub(/[\|:]/, '').tap do |sanitized_id|
+        if sanitized_id.size != id.size
+          Logging::Logger.log "Proxy id '#{id}' changed to '#{sanitized_id}'", :warn
```
bleonard: Minor thing. Might want to say "Makara" here so they have more context.

lib/makara/cookie.rb
bleonard: I see. thanks.

lib/makara/config_parser.rb
rosa: True. For some reason I had assumed without checking that the logger would tag the lines with `[Makara]`, but I realised it was added manually on exception handling. What I have done instead is to add this by default to all lines logged through `Makara::Logging::Logger.log`: https://github.com/taskrabbit/makara/pull/194/commits/51d32a580559893ba8d069e639772473780c9436  What do you think?

bleonard:
@rosa I took some time today to get master to pass Travis. If you want to rebase on that, feel free.

rosa:
> I took some time today to get master to pass Travis. If you want to rebase on that, feel free.

Excellent! 🎉I had looked briefly at the Postgres related failures but only thing I did was to specify the `pg` gem version as `~> 0.18`, which didn't fix the tests locally for me (surely something setup wrongly on my machine). Thanks so much for fixing this!

BTW, I started testing this PR on our app last week, and if everything goes well, we should be running it in production this week 🤞

rosa:
Hey @bleonard, sorry for the delay! Testing this in our app was delayed because of other unrelated infrastructure tasks, but I finally got there last week. We've been using this version of Makara in production for 4 days already, and no issues so far 😄

bleonard:
That's great!
I merged a few more things into master from other PRs. Want to rebase and try that?

bleonard:
I tagged that version as `v0.3.10`
I think your update deserves a semantic version bump at least :-)

rosa:
> I merged a few more things into master from other PRs. Want to rebase and try that?

Done! Basecamp 3 is now running version `v0.3.10` + this pull request in production.

bleonard:
💥 I think I'll release v0.3.10 as-is and then v0.4.0 with this PR.
Do you know of any breaking changes? I think it should be relatively straightforward. To my knowledge, there is no new thing to do.

rosa:
\o/ Thanks! I think the only (obvious) breaking change is that anything depending on the previous context cookie won't work, and that `MemoryStore` has disappeared.

bleonard:
I suppose, though I never considered someone in Ruby or JS reading that cookie. 
We'll see.

There was some Ruby < 2.3 errors with the deprecation string that I fixed on master.
So it and the 0.4.0 tag should be good to go.

I'm about to embark on a pretty serious sabbatical and this is a big change, so I'll let people upgrade to 0.3.10 and/or try this tag out and push this up in May.

rosa:
> There was some Ruby < 2.3 errors with the deprecation string that I fixed on master.

Ouch, sorry about that, and thanks for fixing them 🙏 

> I'm about to embark on a pretty serious sabbatical and this is a big change, so I'll let people upgrade to 0.3.10 and/or try this tag out and push this up in May.

Perfect 👍  Enjoy your sabbatical and thanks so much for all your work and help with this! 🎉

bleonard:
We've been using the v0.4.0 tag for a while and things seems good.
I've pushed it up to ruby gems: https://rubygems.org/gems/makara/versions/0.4.0

Congratulations @rosa - this is a great release. 🎉

rosa:
Thanks @bleonard, and welcome back! We've also been using 0.4.0 since 20th March with no issues 😄 

Thanks again for all your help!

PR #199, opened 2018-03-01
Postgres exists query and travis upgrade

bleonard:


bleonard:
Released as part of 0.3.10: https://rubygems.org/gems/makara/versions/0.3.10

PR #217, opened 2018-08-09
Include more PG Advisory Lock statements to SQL_MASTER_MATCHERS

phil-monroe:
Add different types of PG advisory locks to the list of master only statements.

Here is the list of different advisory locks:
https://www.postgresql.org/docs/current/static/functions-admin.html#FUNCTIONS-ADVISORY-LOCKS-TABLE


phil-monroe:
Hello! I just wanted to follow up to see if this could make its way into master?

PR #223, opened 2018-11-13
Add connection logging in non-Rails enviroments

ankane:
Since connection logging just depends on ActiveRecord, it'd be great to add it even when Rails is not present.

### DIFF
--- a/lib/makara.rb
+++ b/lib/makara.rb
@@ -1,3 +1,4 @@
+require 'active_support'
 require 'makara/version'
 require 'makara/railtie' if defined?(Rails)
 module Makara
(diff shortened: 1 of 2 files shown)


bleonard:
Looks good. Tested in production?

ankane:
No, left my job where we used Makara in production and don't use it for personal projects. Any concerns?

bleonard:
Just the concern that your personal projects are sub-optimally routing their SQL statements 😄 
Or maybe that's what this PR is for. In which case, strong work.

@rosa want to merge this in?

rosa:
@bleonard sure! This LGTM 👍 👍

ankane:
Ha, I'll be ecstatic when my personal projects have enough traffic to use this :) Thank you both 👍

PR #229, opened 2019-03-08
ErrorHandler#harshly no longer fails when backtrace nil

estraph:
When an exception doesn't have a `backtrace` (it is `nil`), the `harshly` method can blow up, masking the underlying exception with this instead:

```
NoMethodError: undefined method `join' for nil:NilClass
```

This PR updates the call to `join` to use the [safe navigation operator](https://en.wikipedia.org/wiki/Safe_navigation_operator#Ruby) which will result in the backtrace in the message being empty, and allowing `harshly` to re-raise the original exception as intended.

### DIFF
--- a/lib/makara/error_handler.rb
+++ b/lib/makara/error_handler.rb
@@ -31,7 +31,7 @@ module Makara
 
 
     def harshly(e)
-      ::Makara::Logging::Logger.log("Harshly handling: #{e}\n#{e.backtrace.join("\n\t")}")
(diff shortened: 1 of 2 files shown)


eugeneius:
Based on `.travis.yml`, it looks like this gem supports Ruby 2.0 and above:

https://github.com/taskrabbit/makara/blob/01195ef2dfd258a53786a4a30a9cbc2aff581e2e/.travis.yml#L19-L26

The safe navigation operator was only added in Ruby 2.3, so we can't use it here.

Issue #151, opened 2017-03-14
makara connection pool management is not thread-safe


aks:
The connection pool management is not thread-safe. So, when running makara in thread-intensive code, such as sidekiq, subtle errors creep in.

The fix is not entirely straightforward: AR 4 and 5 has reorganized the connection pool management to be thread-safe, with lots of mutex usage. Makara has none of that, and it's connection pool management is not thread-safe. 

In fact, makara uses an array of connections for each pool, which is traversed each time a connection decision is being made. Connections are added up to the maximum connection, but there are no protections against simultaneous access and changes across threads. 

In addition, in the master and replica (slave) connection pools, there are connections with different states in each array: blacklisted or not.

We _could_ protect the connection pool array with a semaphore, but traversing an array behind a semaphore is a Bad Idea because it blocks all other threads from traversing that same array. If the semaphore supported read-locking in addition to write-locking then multi-thread traversals would work in parallel, except for adding or removing connections.  

However, IMHO, it would be better to have multiple arrays of connection pools, with each kind of connection being queued into a separate pool (array) of connections. So, blacklisted connections would be maintained in a blacklisted connection pool, and the remaining set of available connections would be in the available connection pool.  Then, each pool could be managed as a thread-safe `queue` object.

Makara should be updated to make best use of the latest AR connection pool code, becoming thread-safe in the process.

bleonard:
Thanks for taking the time to write this up @aks 
It reads like a good summary of the issue at hand.

I don't suppose you want to give it a shot? 😄

aks:
Well, maybe.

I've already forked the makara repo, and have added a hijack on the `connection` method.

However, that change was insufficient.  There are still AR timeouts occurring in our sidekiq processes, which are thread-full, and the stacktraces on those timeouts do not include makara code anywhere.  

Whether or not the timeouts and lack-of-thread-awareness in makara are related is still TBD.

So, there appear to be two issues:
1. makara has incomplete coverage of current AR methods _(it doesn't hijack all the necessary methods)_
2. makara does not manage its connection correctly in a thread-full environment

I've reviewed the ActiveRecord 4.2.6 and 5 code connection adapters and connection handling, to get an idea of how hard it would be to update/rewrite makara to become an alternative connection handler for AR.

In the current AR 4.2.6+ or 5 code, each model has a _connection handler_, which is supposed to determine the appropriate _connection_ using a _connection spec_.  Once a _connection_ is selected, it is cached.

[...]

PR #232, opened 2019-03-26
add sidekiq middleware component to avoid reusing contexts between wo…

pinnymz:
…rkers

### DIFF
--- a/lib/makara.rb
+++ b/lib/makara.rb
@@ -1,6 +1,9 @@
 require 'active_support'
 require 'makara/version'
-require 'makara/railtie' if defined?(Rails)
+if defined?(Rails)
+  require 'makara/railtie'
+  require 'makara/sidekiq/railtie' if defined?(Sidekiq)
+end
 module Makara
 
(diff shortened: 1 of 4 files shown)


pinnymz:
Related to #151, I have confirmed that context reuse between tasks is taking place on recycled threads.  This PR will avoid leaking contexts of previous tasks by clearing out the context as a post-task operation.

sharang-d:
It would be great to get some eyes on this if it fixes the issue with Sidekiq workers!

I'll try to test it in our app where we're seeing `ActiveRecord::ConnectionTimeoutError` exceptions once I can reliably reproduce them locally.

Gargron:
How do you manage to make Sidekiq work with read-replicas at all? Sidekiq jobs will usually fail because they start processing faster than the replica lag so the data the job has to work with will be missing or out of date unless Sidekiq connects directly to master. If you have a solution to that, I'd love to know!

bleonard:
We pointed background workers always to the primary (writer) for this reason and that they often were doing updates anyway.

PR #234, opened 2019-04-16
SET LOCAL is only effective on master

montanalow:
Transactions are not sent to all connections, so `SET LOCAL` is ignored on replicas and causes a warning message.

Reference: https://www.postgresql.org/docs/11/sql-set.html


fuxinhao-crypto:
any progress on this? Is there a chance that someone could give this one a look? currently makara will send `SET LOCAL` to all nodes

fdeschenes:
My logs are filling up with `WARNING:  SET LOCAL can only be used in transaction blocks` because of this issue – would love to see this merged in too.

mlarraz:
Merged in https://github.com/instacart/makara/pull/366

Issue #236, opened 2019-05-09
Question: With stickiness = false is there still stickiness is a single request?


drewB:
If stickiness = false, I know that the same db connection will not be used for subsequent requests but I am unclear what the behavior is during a single request.  

For example:

```
user = User.find(1) #would go to some reader
user.posts.create(:text, "hello") #master
posts = user.posts.all #would this go to master or a reader?
```

So would user.posts.all stick to master or would it grab a reader?

If it does not stick for the request then could we get that behavior by turning stickiness on but setting master_ttl = 0 (or near 0)?

drewB:
I did some testing of this today and this is what I found:

If stickiness = false then it does NOT stick during a single request.  So in my example above `posts = user.posts.all` would go to a reader.

If stickiness = true and master_ttl is set to a very low number (i tried 0.01) then it will stick for the entire request (even if longer than master_ttl).

bleonard:
That's my understanding as well. Stickiness is a concept between _SQL calls_. Setting it to `false` (or using the block) might be right in a situation with known writes that have nothing to do with the reads, for example.

maybe there's a case to be made for `cookie=false` (or some better name) to not do it between requests.

bleonard:
@drewB  thanks for the proactive investigation to your own question!

drewB:
Thanks for confirming what I saw.  

Yes, I think that is where the ambiguity is.  Cause it isn't really between SQL calls (see https://github.com/taskrabbit/makara/issues/206). It is between requests where request #1 is the first one to make a write sql call.  Maybe just an improvement to the README would clear things up.  I will create a PR to try and clear that up.

PR #237, opened 2019-05-09
Add to README more details about stickiness

drewB:
This clears up some ambiguity with how stickiness works.  It is a follow up to https://github.com/taskrabbit/makara/issues/236


bleonard:
thanks! @drewB

PR #246, opened 2019-11-02
Prevent infinite loop when blacklisting disabled

jorgemanrubia:
This prevents a infinite loop situation when blacklisting connections is disabled.

When blacklisting connections are disabled, `Pool#provide` will enter into an infinite loop as long as the connection keeps raising a `BlacklistConnection` error. Because it will never be blacklisted, the current connection will always be available, and it will keep retrying. 

The problem happens when `disable_blacklisting` is `true` and also when `blacklist_duration` is 0.

The patch prevents retrying unless the connection was blacklisted.

It also includes a couple of specs. If you remove the patch and run the specs you will get the infinite loop.

I am not familiar with Makara. I'll be happy to modify the PR if you prefer a different approach.

### DIFF
--- a/makara.gemspec
+++ b/makara.gemspec
@@ -9,7 +9,7 @@ Gem::Specification.new do |gem|
   gem.homepage      = "https://github.com/taskrabbit/makara"
   gem.licenses      = ['MIT']
   gem.metadata      = {
-                        source_code_uri: 'https://github.com/taskrabbit/makara'
+                        'source_code_uri' => 'https://github.com/taskrabbit/makara'
                       }
 
   gem.files         = `git ls-files`.split($\)
--- a/README.md
+++ b/README.md
@@ -132,7 +132,11 @@ There are some edge cases:
 
 ### Errors / blacklisting
 
-Whenever a node fails an operation due to a connection issue, it is blacklisted for the amount of time specified in your database.yml. After that amount of time has passed, the connection will begin receiving queries again. If all slave nodes are blacklisted, the master connection will begin receiving read queries as if it were a slave. Once all nodes are blacklisted the error is raised to the application and all nodes are whitelisted.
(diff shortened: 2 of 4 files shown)


makara.gemspec
```diff
@@ -9,7 +9,7 @@ Gem::Specification.new do |gem|
   gem.homepage      = "https://github.com/taskrabbit/makara"
   gem.licenses      = ['MIT']
   gem.metadata      = {
-                        source_code_uri: 'https://github.com/taskrabbit/makara'
+                        'source_code_uri' => 'https://github.com/taskrabbit/makara'
                       }
```
jorgemanrubia, line 14: bundler was complaining about the keys not being a string ☝️

rosa (commented):
Great catch @jorgemanrubia! I'm not familiar with this part of Makara either, and unsure of what the behaviour should be, but agree that a change is needed for when blacklisting is disabled. The only thing I have doubts about is swallowing the error. By default, errors that get _converted_ into a `Makara::Errors::BlacklistConnection` are [those supposed to be _gracefully_ handled](https://github.com/taskrabbit/makara/blob/dac6be2e01e0511db6715b2b4da65a5490e01cba/lib/active_record/connection_adapters/makara_abstract_adapter.rb#L43-L50), which by default are [connection errors](https://github.com/taskrabbit/makara/blob/dac6be2e01e0511db6715b2b4da65a5490e01cba/lib/active_record/connection_adapters/makara_abstract_adapter.rb#L19-L33) (i.e. DB has gone away, a timeout...). It makes sense to retry in these cases, but only if we blacklist the nodes or we'd be retrying forever as you described. Then, if we aren't blacklisting, and thus not retrying... would it perhaps make sense to let the error bubble up? That is, if blacklisting is disabled, maybe all errors should be _harshly_ handled.

@bleonard do you have more context on what the behaviour should be here?

jorgemanrubia:
Thanks @rosa 🙏, I think you are right: it should raise the error instead of swallowing it when no blacklisting. Still not sure about the whole approach, but if we go with this and don't do that, we would be introducing another kind of serious problem (failing silently). Thanks!

I modified it to raise the `#original_error` wrapped in `Makara::Errors::BlacklistConnection`.

rosa:
@jorgemanrubia what do you think about updating the `README` so this behaviour is more clear?

jorgemanrubia:
Good idea @rosa 👍. I just pushed some changes to the README to clarify blacklisting behavior.

PR #252, opened 2020-03-03
Replace URI.unescape(...) by CGI.unescape(...)

Kevinrob:
With Ruby 2.7, this line raise `warning: URI.unescape is obsolete`.

### DIFF
--- a/lib/makara/config_parser.rb
+++ b/lib/makara/config_parser.rb
@@ -63,7 +63,7 @@ module Makara
       # Converts the given URL to a full connection hash.
       def to_hash
         config = raw_config.reject { |_,value| value.blank? }
-        config.map { |key,value| config[key] = URI.unescape(value) if value.is_a? String }
+        config.map { |key,value| config[key] = CGI.unescape(value) if value.is_a? String }
         config
       end
 


Kevinrob:
ping @bleonard 😊

bleonard:
This stuff always got me. Are these the same thing in all cases?
Is it in all rubies?

Kevinrob:
> This stuff always got me. Are these the same thing in all cases?
> Is it in all rubies?

I just checked the source code. The two functions do the same thing, encoding and pack `%...` values.
`CGI.unescape(...)` exists since at least Ruby 2.1.0. What's the minimum version Makara support?

[Source of CGI.unescape](https://docs.ruby-lang.org/en/2.7.0/CGI/Util.html#method-i-unescape)
[Source of URI.unescape](https://docs.ruby-lang.org/en/2.7.0/URI/Escape.html#method-i-unescape) (which is more difficult to read because it's depend on `DEFAULT_PARSER.unescape`)

ghost:
Is there a timeline on merging this and cutting a new gem release? 🙏

montanalow (approved):
Thanks!

PR #268, opened 2020-09-11
Fix active record connection pool exhaustion

prburgu:


lib/active_record/connection_adapters/makara_abstract_adapter.rb
```diff
@@ -107,10 +107,10 @@ def custom_error_message?(connection, message)
 
 
       hijack_method :execute, :exec_query, :exec_no_cache, :exec_cache, :transaction
-      send_to_all :connect, :reconnect!, :verify!, :clear_cache!, :reset!, :pool=
+      send_to_all :connect, :reconnect!, :verify!, :clear_cache!, :reset!
 
-      control_method :close, :steal!, :expire, :lease, :in_use?, :owner, :schema_cache,
-         :schema_cache=, :_run_checkin_callbacks, :_run_checkout_callbacks, :lock, :seconds_idle
+      control_method :close, :steal!, :expire, :lease, :in_use?, :owner, :schema_cache, :pool=, :pool,
```
nickelser, line 112: might be worth documenting `send_to_all` vs `control_method` and the rationale here

nickelser (approved):
i think it makes sense, but i don't have a good mental model of why this fixes it completely. can you document the changes a bit more in the PR description itself? and, i think a test case that demonstrates this would be good too.

sdhull:
I know it's been awhile but I too would like some insight into what the problem was and how this fixed it

PR #269, opened 2020-09-17
Override equality operator for activerecord connection wrapper

prburgu:
This will fix proxying `==` to raw connection.

```
/var/app/vendor/cache/makara-55759103f219/lib/makara/proxy.rb:137:in `respond_to_missing?'
/var/app/vendor/ruby-2.6.5/lib/ruby/2.6.0/delegate.rb:137:in `respond_to?'
/var/app/vendor/ruby-2.6.5/lib/ruby/2.6.0/delegate.rb:137:in `=='
/var/app/vendor/ruby-2.6.5/lib/ruby/2.6.0/delegate.rb:137:in `=='
/var/app/vendor/bundle/ruby/2.6.0/gems/activerecord-5.2.4.4/lib/active_record/connection_adapters/abstract/connection_pool.rb:568:in `delete'
```


drdrsh (approved):
LGTM.
I did a quick test to see if `any_connection` will be called when `pool.delete conn` is called or not, and your fix works. Thanks for taking care of this.

PR #273, opened 2020-11-02
Add transaction id support to makara_abstract_adapter

elguapo1611:
Always run `select txid_current()` on master without sticking.


thomasfulton (dismissed):

levkk:
You sure this function can be called on a replica to find out what it's current transaction ID is? I was thinking we'd use `pg_current_wal_lsn()`

PR #274, opened 2020-11-23
Lazy Makara

drdrsh:


lib/active_record/connection_adapters/makara_abstract_adapter.rb
```diff
@@ -151,12 +155,16 @@ def initialize(config)
 
       def appropriate_connection(method_name, args, &block)
         if needed_by_all?(method_name, args)
-
-          handling_an_all_execution(method_name) do
-            hijacked do
-              # slave pool must run first.
-              @slave_pool.send_to_all(nil, &block)  # just yields to each con
-              @master_pool.send_to_all(nil, &block) # just yields to each con
+          if ENV["MAKARA_LAZY_CONNECT"] == "true" && method_name == :execute
```
levkk: Wouldn't this enqueue legitimate queries called with `#execute` as well?

drdrsh: `send_to_all` is only called for `SET` queries. Other queries wouldn't hit this codepath.

drdrsh: The definition of needed_by_all looks like this
```
      def needed_by_all?(method_name, args)
        sql = coerce_query_to_sql_string(args.first)
        return true if sql_all_matchers.any?{|m| sql =~ m }
        false
      end
```
And `sql_all_matchers` will only match `SET` queries. So this code path will only be hit by `SET` queries.

lib/makara/connection_wrapper.rb
```diff
@@ -97,12 +103,24 @@ def execute(*args)
           args[0] = replace
         end
       end
+      run_queue if ENV["MAKARA_LAZY_CONNECT"] == "true"
```
levkk: I think instead of queuing, you could run initialize here if it has not been initialized already.

drdrsh: I think we will only have a `ConnectionWrapper` object if the pool is already initialized. Before initialization we only have a pool object without any wrapped connections.

PR #275, opened 2020-12-22
Set up automatic publishing to Github and Rubygems

mlarraz:


.github/workflows/gem-push.yml
```diff
@@ -0,0 +1,33 @@
...
+      run: |
+        mkdir -p $HOME/.gem
+        touch $HOME/.gem/credentials
+        chmod 0600 $HOME/.gem/credentials
+        printf -- "---\n:github: ${GEM_HOST_API_KEY}\n" > $HOME/.gem/credentials
+        gem push --verbose --key github --host https://rubygems.pkg.github.com/${OWNER} *.gem
+      env:
+        GEM_HOST_API_KEY: "Bearer ${{secrets.GITHUB_TOKEN}}"
+        OWNER: ${{ github.repository_owner }}
+    - name: Publish to RubyGems
```
scashin133: Are there going to be cases where we only want to publish to github? When would we want to publish to rubygems and github?

mlarraz: This assumes we'll just publish everything to both, as a redundancy.

scashin133: For a lot of these gems though we explicitly DON'T want them to be public right? otherwise we wouldn't need the private github repo. I'd prefer renaming this workflow to something that makes that a lot more obvious or just not include the public aspect at all until we have a usecase for it.

mlarraz: Github Packages follows the repo's privacy settings. Since this is a public repo the Github package will also be public. Not strictly necessary but might be nice to have

scashin133: okay so this is for repos like amountable that are public? I would prefer that we make that more explicit in the workflow name otherwise I think someone is going to make a mistake

mlarraz: Updated to clarify. We can also guard against accidental publishes by whitelisting `allowed_push_host` in a gemspec's metadata, which is more resilient.

scashin133 (approved):
Thanks for making the changes!

PR #283, opened 2021-01-14
Fix on Ruby 3 

mlarraz:
Because of how much method delegation and metaprogramming this gem uses, there were a number of issues due to the Ruby 3 keyword argument change.

Most are resolvable by adding `ruby2_keywords`, but the stuff using `instance_eval` is a bit more involved.

### DIFF
--- a/lib/makara/connection_wrapper.rb
+++ b/lib/makara/connection_wrapper.rb
@@ -103,12 +103,10 @@ module Makara
 
     # we want to forward all private methods, since we could have kicked out from a private scenario
     def method_missing(m, *args, &block)
-      if _makara_connection.respond_to?(m)
-        _makara_connection.public_send(m, *args, &block)
-      else # probably private method
-        _makara_connection.__send__(m, *args, &block)
-      end
+      _makara_connection.send(m, *args, &block)
     end
+
+    ruby2_keywords :method_missing if Module.private_method_defined?(:ruby2_keywords)
 
     def respond_to_missing?(m, include_private = false)
       _makara_connection.respond_to?(m, true)
@@ -120,7 +118,7 @@ module Makara
     # all extra functionality is in the format of _makara*
     def _makara_decorate_connection(con)
 
-      extension = %Q{
+      extension = <<~RUBY
         # the proxy object controlling this connection
         def _makara
           @_makara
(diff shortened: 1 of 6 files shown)


lib/makara/connection_wrapper.rb
```diff
@@ -144,38 +142,44 @@ def _makara_hijack
...
+      RUBY
+
+      args = RUBY_VERSION >= "3.0.0" ? "..." : "*args, &block"
 
       # Each method the Makara::Proxy needs to hijack should be redefined in the underlying connection.
       # The new definition should allow for the proxy to intercept the invocation if required.
       @proxy.class.hijack_methods.each do |meth|
-        extension << %Q{
-          def #{meth}(*args, &block)
+        method_call = RUBY_VERSION >= "3.0.0" ? "public_send(#{meth.inspect}, ...)" : "#{meth}(*args, &block)"
```
mlarraz, line 152: We need `public_send` here because `foo.bar=(...)` is not valid syntax in Ruby 3.0 (even though `foo.bar(...)` is)

spec/active_record/connection_adapters/makara_postgresql_adapter_spec.rb
```diff
@@ -73,7 +73,13 @@
       Test::User.exists? # flush other (schema) things that need to happen
 
       con = connection.replica_pool.connections.first
-      expect(con).to receive(:exec_query).with(/SELECT\s+1\s*(AS one)?\s+FROM .?users.?\s+LIMIT\s+.?1/, any_args).once.and_call_original
+
+      expect(con).to receive(:exec_query) do |query|
```
scashin133, line 77: Seems like this change is causing a failure for this matrix run: https://github.com/instacart/makara/pull/283/checks?check_run_id=1730065278

mlarraz: No, this spec flaps on jruby for some reason. It's the same on master now.

mlarraz: See https://github.com/instacart/makara/runs/1705529911

scashin133 (approved):

lib/makara/proxy.rb
```diff
@@ -118,20 +124,20 @@ def strategy_class_for(strategy_name)
...
+        return super
       end
 
       any_connection do |con|
-        if con.respond_to?(m)
-          con.public_send(m, *args, &block)
-        elsif con.respond_to?(m, true)
-          con.__send__(m, *args, &block)
+        if con.respond_to?(m, true)
+          con.send(m, *args, &block)
```
jeremy, line 132: Subtle regression here. We'll now call a private `#foo` method on the ConnectionWrapper instead of calling the public `#foo` method on the wrapped ConnectionAdapter. That's because using `#send` will dispatch the private method immediately whereas `#public_send` will fall back to method_missing, which is where the Makara ConnectionWrapper talks to the AR ConnectionWrapper.

Example
```ruby
c.send(:select, "select 1")
*** TypeError Exception: wrong argument type String (expected Array)

c.public_send(:select, "select 1")
   (2.1ms)  select 1
#<ActiveRecord::Result:0x00007f8784b4e7f0 @columns=["1"], @rows=[[1]], @hash_rows=nil, @column_types={}>
```

This is the cause of #326

mlarraz: Thanks for catching this. TIL about `#send` not hitting `#method_missing`.

If you can send a PR and maybe add a spec to prevent a regression I'll get it merged ASAP.

jeremy: Sure thing. Thanks @mlarraz.

PR #284, opened 2021-01-14
Remove checks for unsupported versions

mlarraz:


spec/active_record/connection_adapters/makara_postgis_adapter_spec.rb
```diff
@@ -1,16 +1,12 @@
 # RGeo doesn't play well with JRuby and to avoid complicated test setup
 # we're only testing ActiveRecord version ~> 4.2
-# also rgeo only works on 2.1+
 
-rmajor, rminor, rpatch = RUBY_VERSION.split(/[^\d]/)[0..2].map(&:to_i)
 require 'active_record'
 
 # TODO: test this in AR 5+ ?
```
mlarraz, line 6: These are still failing on AR 5+, will address separately

scashin133 (approved):

PR #286, opened 2021-01-14
Deprecate 'slave' in favor of 'replica'

mlarraz:
Replica is a clearer term to use.

Old configs and methods should still work but will trigger deprecation warnings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 All notable changes to this project will be documented in this file.
 
 ## Unreleased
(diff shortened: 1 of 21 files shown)


scashin133 (commented):
I think this makes sense. We should also update `master` to `primary`, were you planning on doing that in a follow-up PR?

mlarraz:
> I think this makes sense. We should also update `master` to `primary`, were you planning on doing that in a follow-up PR?

Yeah that is a little more involved than this change

scashin133 (approved):

PR #290, opened 2021-01-19
Deprecate 'master' in favor of 'primary'

mlarraz:
Also deprecates the use of the `slave` role (although this doesn't actually matter since we just assume anything but `primary` is a replica).


lib/makara/config_parser.rb
```diff
@@ -158,29 +169,45 @@ def id
...
       replica_configs
     end
 
     protected
 
     def all_configs
-      @makara_config[:connections].map do |connection|
-        base_config.merge(makara_config.except(:connections))
-                   .merge(connection.symbolize_keys)
+      @all_configs ||= @makara_config[:connections].map do |connection|
```
scashin133, line 197: This seems like a behavior change. Now we won't be re-calculating all_configs each time it is called. That may ultimately be desired behavior but wanted to call out that potentially this could break things

lib/makara/proxy.rb
```diff
@@ -205,52 +210,57 @@ def appropriate_connection(method_name, args)
...
-    def needs_master?(method_name, args)
-      true
+    # Do these args require a primary connection
+    def needs_primary?(method_name, args)
+      if respond_to?(:needs_master?)
+        warn "#{self.class}#needs_master? is deprecated. Switch to #needs_primary?"
+        needs_master?(method_name, args)
+      else
+        true
+      end
```
scashin133, line 263: I don't fully understand this change. How could it respond_to `needs_master?` since your removed that method? It also seems like it would be throwing a deprecation warning potentially when using the correct method. I'm sure there is some method magic happening that I just don't see but wanted to call it out.

scashin133 (commented):
Had a couple of questions about some of the changes that appear to either be changing the behavior or don't visibly make sense. If you have good reasons I'm happy to approve.

lib/makara/proxy.rb
mlarraz: My understanding is this method is meant to be overwritten in subclasses by users. So if someone has defined their own `needs_master?`, we should still call it but emit this warning that they should rename the method.

lib/makara/config_parser.rb
mlarraz: AFAIK these are immutable structures anyway (since they're just loaded from config files)

scashin133 (approved):

PR #338, opened 2021-11-13
Adds the ability to retry for the pool of connection if specified in makara config

vishnun:
This is an attempt to do graceful failover by re-trying at the lowest level.
(This isn't ready to be merged in master for makara but we'll use this branch)

Example config:

```
development:
  database: speckel_development
  adapter: postgresql_makara
  host: localhost
  username: speckel
  password: 23432
  pool: 20
  makara:
    id: 'development-app-postgres'
    disable_blacklist: true
    master_ttl: 0.3
    sticky: true
    retry_exceptions:
      - name: ActiveRecord::StatementInvalid
        retry_count: 5
        time_between_retries_in_seconds: 0.2
    connections:
      - role: master
        name: DB-Writer
        host: localhost
        port: 5432
        blacklist_duration: 0
      - role: slave
        name: DB-Reader
        host: localhost
        port: 5433

```

### DIFF
--- a/lib/makara/proxy.rb
+++ b/lib/makara/proxy.rb
@@ -44,42 +44,6 @@ module Makara
           end
         end
       end
-
-      private
-
-      def _execute_with_connection_and_retry_exceptions(retry_exceptions, args, block, method_name)
-        begin
-          should_retry = false
-          potentially_stale_connection = nil
-
-          appropriate_connection(method_name, args) do |con|
-            con.send(method_name, *args, &block)
-            potentially_stale_connection = con
-          end
-
-        rescue Exception => actual_exception
-          retry_attempts = retry_exceptions.inject({}) { |memo, retry_exception| memo[retry_exception['name']] = 0; memo }
-          retry_exceptions.each do |retry_exception|
-            if actual_exception.class.to_s == retry_exception['name']
-              potentially_stale_connection.disconnect!
-              sleep retry_exception['time_between_retries_in_seconds'] || 0.1
-              retry_attempt = (retry_attempts[actual_exception.class.to_s] += 1)
-
-              if retry_attempt < retry_exception['retry_count'] && retry_attempt < MAX_RETRY_ATTEMPTS
-                should_retry = true
-              end
-            end
-          end
-          retry if should_retry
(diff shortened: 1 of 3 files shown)


lib/makara/proxy.rb
```diff
@@ -162,6 +166,43 @@ def any_connection
...
+
+      rescue Exception => actual_exception
+        retry_attempts = retry_exceptions.inject({}) { |memo, retry_exception| memo[retry_exception['name']] = 0; memo }
+        retry_exceptions.each do |retry_exception|
+          if actual_exception.class.to_s == retry_exception['name']
+            potentially_stale_connection.disconnect!
+            sleep retry_exception['time_between_retries_in_seconds'] || 0.1
+            retry_attempt = (retry_attempts[actual_exception.class.to_s] += 1)
+
+            max_attempts = retry_exception['retry_count'] || MAX_RETRY_COUNT
```
iain-bryson, line 186: Better name would be `DEFAULT_RETRY_COUNT` since it's not a maximum unless you don't specify one

lib/makara/proxy.rb
```diff
@@ -162,6 +166,43 @@ def any_connection
...
+          con.send(method_name, *args, &block)
+          potentially_stale_connection = con
+        end
+
+      rescue Exception => actual_exception
+        retry_attempts = retry_exceptions.inject({}) { |memo, retry_exception| memo[retry_exception['name']] = 0; memo }
+        retry_exceptions.each do |retry_exception|
+          if actual_exception.class.to_s == retry_exception['name']
+            potentially_stale_connection.disconnect!
+            sleep retry_exception['time_between_retries_in_seconds'] || 0.1
```
iain-bryson, line 183: I wonder if a bit of randomization here would be useful to avoid a load spike after the instance comes up.

iain-bryson (dismissed):
LGTM some comments that are completely optional and I don't think are obviously worth resetting testing.

lib/makara/proxy.rb
vishnun: It is a MAX count as well if you see the condition below :)

PR #353, opened 2022-05-02
Use new delegation syntax for ruby >= 2.7

rahul342:
Delegation syntax is supported in ruby 2.7 and above 
https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/

This will help with not issuing deprecation warning when running ruby >= 2.7

### DIFF
--- a/lib/makara/connection_wrapper.rb
+++ b/lib/makara/connection_wrapper.rb
@@ -142,12 +142,12 @@ module Makara
         end
       RUBY
 
-      args = RUBY_VERSION >= "3.0.0" ? "..." : "*args, &block"
+      args = RUBY_VERSION >= "2.7" ? "..." : "*args, &block"
 
       # Each method the Makara::Proxy needs to hijack should be redefined in the underlying connection.
       # The new definition should allow for the proxy to intercept the invocation if required.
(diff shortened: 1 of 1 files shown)


rahul342:
cc @mlarraz (author of original commit) hope it's ok to tag.

mlarraz:
I forgot why I set this to 3.0 and not 2.7. It's possible there was some difference in the syntax between the versions. These tests are not comprehensive so additional verification would be good to have here.

rahul342:
@mlarraz Can you elaborate what additional verification? I tested this change in our repo with Rails 6.0 and ruby 2.7, seems to be working fine.
Possibly tag another reviewer for an extra pair of eyes?

mlarraz:
See https://eregon.me/blog/2021/02/13/correct-delegation-in-ruby-2-27-3.html
> There is the def foo(...); target(...); end syntax but it fails to parse on Ruby <= 2.6, and **only works with no other arguments at all in Ruby 2.7.0 - 2.7.2**

and this related issue https://bugs.ruby-lang.org/issues/16378

rahul342:
Thanks! Will close this one for now.

PR #355, opened 2022-08-04
Refer connection payload in addition to connection_id

yyamanoi1222:
https://github.com/rails/rails/commit/d140ab073f296a2a689a4dbd58cf9434c6b2a484

<code>connection_id</code> is removed from <code>sql.active_record</code>notification in this commit.
So It was no longer showing up in the log in Rails6.1

I fixed it by referring to <code>event.payload[:connection]</code> in addition to <code>event.payload[:connection_id]</code>


mlarraz:
Merged in https://github.com/instacart/makara/pull/367

PR #357, opened 2022-08-22
[Fix] 356 Issue replica instead of primary

hitendrasingh:
Fix for this issue.

https://github.com/instacart/makara/issues/356


17jmumford (approved):

mlarraz:
Fixed in https://github.com/instacart/makara/pull/364

PR #365, opened 2023-06-23
Test on Rails 7

mlarraz:


toregeschliman (approved):

yaroslav-cc:
Hi @mlarraz . Thanks for this change. Do you happen to know if there's a new version release planned soon that would include this (and everything else in master) in it?

PR #366, opened 2023-06-23
Skip SET LOCAL

mlarraz:
SET LOCAL is only effective on master

Transactions are not sent to all connections, so "SET LOCAL" is ignored on replicas and causes a warning message.

Taken from https://github.com/instacart/makara/pull/234


toregeschliman (approved):

kakubin:
@mlarraz 
Could you release new version?

Issue #382, opened 2023-11-05
Deprecation warning with Rails 7.1


jwoodrow:
I managed to get makara working (at least it seems to be working so far) but I've got a pesky warning

> DEPRECATION WARNING: new is deprecated and will be removed from Rails 7.2 (use ActiveRecord::ConnectionAdapters::SchemaReflection instead)

caused by this line `lib/active_record/connection_adapters/makara_abstract_adapter.rb:193`

I have no real idea how the makara magic works but I noticed there wasn't an issue opened with this problem, thanks for the hard work 🙏 

Also out of curiosity, I can't seem to get automatic failover to master working in a specific case, I want to failover on `ActiveRecord::RecordNotFound` exception to take into account replication lag in jobs (I configured my jobs to not stick to primary so I could prioritize using the replica database as much as possible until they do a write but some jobs are enqueued right after a create to the database and for those I would like to be able to fallback to the primary database)

jwoodrow:
Did some blind digging and testing to try and remove the warning, this seems to work in place of the deprecated call

```ruby
schema_reflection = ActiveRecord::ConnectionAdapters::SchemaReflection.new(nil).load!(@proxy)
@schema_cache = ActiveRecord::ConnectionAdapters::BoundSchemaReflection.new(schema_reflection, @proxy)
```

Wouldn't mind if someone more knowledgable could confirm if this is sufficient or not 🤔

PR #384, opened 2023-11-12
Use `SchemaReflection` and `BoundSchemaReflection` instead of `SchemaCache`

jwoodrow:
Rails 7.1 warns that `ActiveRecord::ConnectionAdapters::SchemaCache` will be removed in Rails 7.2.

I tried to look around at what was being returned by `ActiveRecord::ConnectionAdapters::SchemaCache` and try to rebuild the exact same class and this seems to work out.

Ran the CI on our end and it seems to work if I use the `ActiveRecord` version as a condition for the switch. There's most-likely a better way of doing this if you have suggestions.

---

Solves #382

### DIFF
--- a/lib/active_record/connection_adapters/makara_abstract_adapter.rb
+++ b/lib/active_record/connection_adapters/makara_abstract_adapter.rb
@@ -212,8 +212,12 @@ module ActiveRecord
           @proxy = proxy
           @owner = nil
           @pool = nil
-          schema_reflection = ActiveRecord::ConnectionAdapters::SchemaReflection.new(nil).load!(@proxy)
-          @schema_cache = ActiveRecord::ConnectionAdapters::BoundSchemaReflection.new(schema_reflection, @proxy)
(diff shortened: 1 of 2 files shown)


lib/active_record/connection_adapters/makara_abstract_adapter.rb
```diff
@@ -212,7 +212,12 @@ def initialize(proxy)
           @proxy = proxy
           @owner = nil
           @pool = nil
-          @schema_cache = ActiveRecord::ConnectionAdapters::SchemaCache.new @proxy
+          if ActiveRecord.version >= Gem::Version.new('7.1.0')
+            schema_reflection = ActiveRecord::ConnectionAdapters::SchemaReflection.new(nil).load!(@proxy)
```
moustafasallam, line 216: This still not going to work because you are passing `nil` to `ActiveRecord::ConnectionAdapters::SchemaReflection.new`

The right approach would be passing the schema_cache_path to `ActiveRecord::ConnectionAdapters::SchemaReflection.new` in order to fix it

